PR #584, opened 2013-12-13
enable pdftex extensions with \usepackage{microtype}

Shark64:
I've tried to enable microtype and seems to work well (fewer overfull hboxes).
The only problem was hyphenation of 2 section titles, but disabling tracking on those 2 pages seems to fix it.


peterlefanulumsdaine:
I’ve asked [a question about this](http://tex.stackexchange.com/questions/337149/what-typographic-warts-can-arise-from-adding-microtype-to-a-project) on tex.stackexchange — asking what we should be checking for after adding microtype.

In any case: it seems to me more trouble than it’s worth to add `microtype` to the current edition, since it would require re-proofing the whole book.  But for the second edition (if that happens), the book will need to be re-proofed anyway, so there’s presumably no reason not to add it then.  So I guess we should keep this PR around, but labelled as “for 2nd ed”?

mikeshulman:
Makes sense to me.

Issue #601, opened 2014-01-11
How to insert new theorems in edition 1


andrejbauer:
The incremental improvements to the book are working very nicely, I think. But the fact that we decided not to change numbering is making it hard to iterate towards a second edition. In fact, it is going to be hard to ever begin working on a second edition because the incremental corrections work so well.

Of course, not changing the numbering was the correct decision to make. People refer to theorems by their numbers. However, let us think about relaxing the scheme so that we can in fact add new theorems. It is simple, really, once we recognize the fact that nothing prevents us from using a dense linear order.

For example, we could insert new theorems in such a way that they would receive numbers between those already in the book. I think it could be made to work. At some point we can "reset" the scheme and announce a 2nd edition, but I suspect it would be better to plan for a neverending stream of improvements.

mikeshulman:
I don't see why the sequence in which theorems were added should be relevant.  The ordinary reader shouldn't care about that.  The ideal world for the reader would be to have all the theorems numbered with sequential positive integers; the only reason we need fractions at all is that we don't want to change the numbering of existing theorems when adding new ones in between two old ones.

(FWIW, at this point I think it's unlikely that anything of this sort will ever be done.  We've made it through 10+ years of errata without ever needing to add a new theorem in between two existing ones.)

Bolpat:
> I don't see why the sequence in which theorems were added should be relevant.

It’s only vaguely relevant if you read a newer version and wonder: “Hey, I don’t remember this thing!?” And if its number is 4.2.10⅓, it might tell you that it was added later. There might be some value in knowing, of A, B, C, D, and E, that A and E came first and C was added later, and B and D were added after C. There’s very little value knowing in which order B and D were added.

PR #613, opened 2014-01-16
fractional theorem numbers; see #601

mikeshulman:


macros.tex
```diff
@@ -676,18 +676,77 @@
...
+\newcount\@xone
+\newcount\@xtwo
+\def\checknewthmfrac#1#2{
+  \ifnum0<#1\else\errmessage{Fractional theorem numbers must be greater than zero}\fi
+  \ifnum#1<#2\else\errmessage{Fractional theorem numbers must be less than one}\fi
+  \@xone=#1
+  \@xtwo=#2
+  \multiply\@xtwo by \value{@newthmnum}
+  \multiply\@xone by \value{@newthmdenom}
+  \ifnum\@xone>\@xtwo\else\errmessage{Theorem numbers out of order}\fi
```
JasonGross, line 745: I don't know if you ever plan on merging this, but if you do, it might be nice to give expect errors like
```suggestion
  \ifnum\@xone>\@xtwo\else\errmessage{Theorem numbers out of order: #1/#2 <= \arabic{@newthmnum}
  / \arabic{@newthmdenom}}\fi
```
(Or maybe we want `\the\value` instead of `\arabic`?)

PR #728, opened 2014-10-20
add an explicit topological example to remark 1.12.1

mikeshulman:
As suggested by @martinescardo on the mailing list


mikeshulman:
bump

andrejbauer:
Other than the misleading idea that a path is a map from an interval, the example is probably helpful.

PR #751, opened 2014-12-24
Fixed several tiny (non-mathematical) typos

iblech:
This pull request corrects several tiny typos. They are all non-mathematical, so probably do not need an entry in the errata list.

I added a few non-breaking spaces, only in places where it looked really necessary to do so in my particular printing (A4).

Merry Christmas!


iblech:
Thanks for the feedback! I added a further commit to this pull request. The two URLs in front.tex are now displayed using a fixed-width font (\url{http://...}). Please let me know whether I should make more amends to this pull request.

mikeshulman:
Sorry to be so picky, but let's keep them as hyperlinks too, something like `\href{http://...}{\url{http://...}}`?

pthariensflame:
@mikeshulman Why not make this pattern an actual command?

mikeshulman:
@pthariensflame If we were going to use it more than twice in the whole book, maybe...

iblech:
No need to be sorry! But I think using \url is enough: In addition to rendering the link in a fixed-width font, \url registers the link as a clickable hyperlink. Or does it not work for you?

mikeshulman:
Oh, hm, yes.  Sorry!

iblech:
Great! Thanks for the review. :-)

PR #757, opened 2015-01-05
P in the flattening lemma is defined *recursively*, not inductively

mikeshulman:
(unless I'm crazy)


andrejbauer:
Oh, we could split hairs here since recursion is a non-dependent induction.

mikeshulman:
Hmm, to me "X is defined inductively" means that X is an inductive type, not that X is a dependent function defined using the induction principle of its domain.

andrejbauer:
Ah, this never occurred to me because I looked at `P : W -> Type` and thought immediately of dependent vs. non-dependent function issue. A type family *is* a function.

mikeshulman:
Of course, but there are also inductive families.

PR #761, opened 2015-01-31
what the book is not

awodey:
revised introduction slightly


andrejbauer:
Thanks, looks good to me.

PR #762, opened 2015-02-01
Fix a few things about the surreal numbers

mikeshulman:
While formalizing the surreal numbers, I noticed some typos and an omission from the definition of addition.

### DIFF
--- a/errata.tex
+++ b/errata.tex
@@ -737,6 +737,10 @@ While the page numbering may differ between copies with different version marker
   & 222-g3453cf1
   & This is the intermediate value theorem, not the mean value theorem.\\
   %
+  \cref{eg:surreal-addition}
+  & % merge of c66efca96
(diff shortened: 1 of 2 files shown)


andrejbauer:
You're formalizing the surreal numbers? Wow, how painful is it?

mikeshulman:
It's not as bad as you might think; I submitted a PR to the HoTT Coq repo.

mikeshulman:
bump

mikeshulman:
bump again.  I'd especially like to get this merged before updating the posted version since it is an actual mathematical error.

andrejbauer:
Must it be me who digs into this and proof reads the changes?

EgbertRijke:
I'm not yet at the changes, but I found a typo of the form "which it is" on line 2725.

I'm reading the whole thing.

mikeshulman:
I don't think "which it is" was a typo.

> We will also need the joint recursion principle, \define{$(\NO,\le,<)$-recursion}, which it is convenient to state as follows.

I.e. we will need the joint recursion principle, and it is convenient to state the joint recursion principle as follows.  But if that's unclear, it could probably be rephrased somehow.

mikeshulman:
Thanks for reading it!!

mikeshulman:
I've fixed all your suggestions.  Thanks for reading it!!

PR #767, opened 2015-02-15
Typos

Pierre-Yves-Gaillard:

PR #768, opened 2015-02-15
Typos

Pierre-Yves-Gaillard:


equivalences.tex
```diff
@@ -87,8 +87,8 @@ \section{Quasi-inverses}
   Let $g:\prd{x:A} \brck{a=x}$ be as given by~\ref{item:autohtpy2}.  First we
   observe that each type $\id[A]xy$ is a set.  For since being a set is a mere
   proposition, we may apply the induction principle of propositional truncation, and assume that $g(x)=\bproj
-  p$ and $g(y)=\bproj q$ for $p:a=x$ and $q:a=y$.  In this case, composing with
-  $p$ and $\opp{q}$ yields an equivalence $\eqv{(x=y)}{(a=a)}$.  But $(a=a)$ is
+  p$ and $g(y)=\bproj{p'}$ for $p:a=x$ and $p':a=y$.  In this case, composing with
```
mikeshulman: What's wrong with `q`?

equivalences.tex
```diff
@@ -448,7 +448,7 @@ \section{Contractible fibers}
 For any $f:A\to B$ we have ${\iscontr(f)} \to {\ishae(f)}$.
 \end{thm}
 \begin{proof}
-Let $P : \iscontr(f)$. We define an inverse mapping $g : B \to A$ by sending each $y : B$ to the center of contraction of the fiber at $y$:
+Let $P : \iscontr(f)$. We define a quasi-inverse mapping $g : B \to A$ by sending each $y : B$ to the center of contraction of the fiber at $y$:
```
mikeshulman: Same comment here as on #767 .

equivalences.tex
Pierre-Yves-Gaillard: The variable q occurs a lot of times in the statement of Lemma 4.1.2 and in its proof. It seems to me that the first three occurrences of q in the proof should be changed to a name of variable which doesn't appear elsewhere in the proof. Note for instance that we have q:a=a in the statement, and q:a=y at the beginning of the proof.

equivalences.tex
Pierre-Yves-Gaillard: I agree again.

equivalences.tex
mikeshulman: Makes sense, thanks for explaining.

Pierre-Yves-Gaillard:
Replaced with PR #770.

PR #773, opened 2015-02-26
Made some changes to the Introduction

vladimirias:
I have corrected some places on the first two or three pages of the introduction. Hopefully I removed some things that were wrong without adding new things that are wrong...

PR #775, opened 2015-03-02
fix reference to Church in introduction

mikeshulman:
This was the most recent proposal to fix one of the issues raised at #773.

### DIFF
--- a/introduction.tex
+++ b/introduction.tex
@@ -33,7 +33,9 @@ The ultimate theory will almost certainly not look exactly like the one describe
 
 \subsection*{Type theory}
 
-Type theory was originally invented by Bertrand Russell \cite{Russell:1908},\index{Russell, Bertrand} as a device for blocking the paradoxes in the logical foundations of mathematics  that were under investigation at the time. It was later developed as a rigorous formal system in its own right (under the name ``$\lambda$-calculus'') by Alonzo Church \cite{Church:1933cl,Church:1940tu,Church:1941tc}.  Although it is not generally regarded as the foundation for classical mathematics, set theory being more customary, type theory still has numerous applications, especially in computer science and the theory of programming languages~\cite{Pierce-TAPL}.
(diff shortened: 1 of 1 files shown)


vladimirias:
I disagree with this correction strongly. Russell type theory has not become a rigorous formal system under the name typed lambda calculus. This is materially wrong.

mikeshulman:
Great, let's continue the discussion here, civilly and politely.  I just wanted to give us a place to do that separately from the morass of #773.

The suggestion does not say that Russell's type theory *became* blah blah, but that it *was eventually developed into* blah blah.  I think that's vague enough to account for the facts, but we could I guess make it even vaguer by saying something like "The idea of type theory was eventually developed into...".

vladimirias:
What’s wrong with my suggestion?

***
This was followed by a period of development culminating with the paper by Alonzo Church \cite{Church:1940tu} where Church connected type theory with $\lambda$-calculus. With small modifications Church's type theory survives today under the name ``higher-order logic”.
***

This is much more informative explaining to the reader that Church’s 1940 paper introduces that which is today known as HOL. It is important and useful information.

mikeshulman:
I'm not sure bringing in HOL is useful for the book at this point, though; our goal is just to give enough of the history to bring us to the point of the type theory we care about in the book, plus it could be misleading because Church didn't use that phrase.  Also I'm not sure whether or not the word "culminating" is appropriate, as it could be read to imply that the subject stopped after that.  What about something even vaguer like "This was developed further by many people over the next few decades; a notable milestone was the work of Church~[...] connecting type theory to lambda-calculus."?

awodey:
Typed lambda-calculus is still a thing: http://www.mimuw.edu.pl/tlca/

mikeshulman:
I updated the PR to (more or less) my most recent proposal, which no one objected to yet.

EgbertRijke:
This pull request was made a while ago, and the change seemed reasonable. If someone wants to refer to Church in a different way, it can be proposed in a new pull request.

PR #776, opened 2015-03-02
improve wording in intro about the simplicial interpretation

mikeshulman:
This is a proposal to fix one of the other wording problems raised at #773.  I didn't fully understand the objection to the existing sentence, though, so I may not have actually fixed the problem; please let me know.

### DIFF
--- a/introduction.tex
+++ b/introduction.tex
@@ -114,7 +114,8 @@ The original semantic models of Awodey--Warren and Voevodsky use well-known noti
 \index{Quillen model category}%
 \index{model category}%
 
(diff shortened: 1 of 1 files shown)


vladimirias:
LGTM

andrejbauer:
Look good. Do we need to put in references to Church and Hofmann & Streicher? It's not a research paper, and we're in the introduction, so too many references can be off-putting.

mikeshulman:
I don't think we *need* to, certainly not in this PR.  There are citations to both of them elsewhere in the intro, I think.

mikeshulman:
There were two acks and no disagreement, so I'm going to merge this.

PR #782, opened 2015-04-04
Added a missing period to logic.tex, line 204

Pierre-Yves-Gaillard:
Added a missing period to logic.tex, line 204.


mikeshulman:
I'm not sure I like the idea of being tedious about always insisting that "an equivalence" is a pair; the coercion from an equivalence to its underlying function ought usually to be implicit.

Pierre-Yves-Gaillard:
Ok. I tried to update the pull request (I'm far from mastering GithHub...)

mikeshulman:
Should we emphasize the implicitness of this coercion more somewhere?

Pierre-Yves-Gaillard:
To avoid any misunderstanding: I suggested to change Line 49 of equivalences.tex because I was not completely happy with the following phrases in the second and third sentences of the proof of Lemma 4.1.1:

"... we may assume that (f,e) is of the form idtoeqv(p) for some p:A=B. Then ... we may assume p is refl_A, in which case idtoeqv(p) is id_A".

PR #786, opened 2015-04-16
Mention a higher coherence path, close #785

### LINKED ISSUE #785: Lack of examples of higher level relations in Section 2.6
In the first para of Section 2.5 the book says, "Our goal in the rest of the chapter is to make explicit how this higher structure behaves in the case of the particular types defined in Chapter 1."  However, all the equations in Section 2.6 on Cartesian products are of level 1, that is, dealing with paths between points.  I don't know if this is the case with the remaining sections of Chapter 2 also.

It'd be good if the book can give at least one example of a higher level relation in Section 2.6, and similarly for the other sections of Chapter 2.  For instance, what is the relation between the homotopies of Lemma 2.1.4 for AxB, and the same homotopies for A and for B?  I think it's a misnomer to call Section 2.5 "The higher groupoid structure of type formers", and then give examples of only the the lowest level of relations in the following sections.

@pthariensflame Thanks for the pointer about editing.  I've corrected the typos here, and removed the two comments I'd used earlier to correct them.

mikeshulman:


basics.tex
```diff
@@ -1316,8 +1316,29 @@ \section{Cartesian product types}
...
+\begin{alignat*}{2}
+  \projpath{i}(\refl{(z : A \times B)}) &= \refl{\proj{i} z} &\qquad (i=1,2)\\
+  \pairpath(\opp p, \opp q) &= \opp{\pairpath(p,q)}\\
+  \pairpath(p\ct q, p'\ct q') &= \pairpath(p,p') \ct \pairpath(q,q').
+\end{alignat*}
 All of these equations can be derived by using path induction on the given paths and then returning reflexivity.
+The same is true for the rest of the higher groupoid structure considered in \cref{sec:equality}, although it begins to get tedious to insert enough other coherence paths to yield an equation that will typecheck.
+For instance, if we denote the path in \cref{thm:omg}\ref{item:omg4} by $\ctassoc(p,q,r)$ and the last path displayed above by $\pairct(p,q,p',q')$, then for any $u,v,z,w:A\times B$ and $p,q,r,p',q',r'$ of appropriate types we have
+\begin{equation*}
+  \begin{array}{l}
```
andrejbauer: Any reason why this isn't done using `multline` or `align` or whichever evnironment `amsmath` provides?

mikeshulman: I thought it would look better to align the `\ct` symbols on each side of the equation separately, so the reader can easily separate the six ingredient paths.  But I'm not attached to it; if you can cook up something using an ams environment that looks better, be my guest!

mikeshulman: Are you going to try to make this look nicer, or can we go ahead and merge this as-is (or does it look so bad that that's not okay)?

PR #792, opened 2015-06-01
Replace pi notation for non-dependent functions

11Kilobytes:
The style convention that ∏(x:A)B should be replaced with A -> B, if B is not dependent on x is not followed in the proof for associativity of concatenation. This commit fixes this.

### DIFF
--- a/basics.tex
+++ b/basics.tex
@@ -466,7 +466,7 @@ It is a familiar fact in topology that when we concatenate a path $p$ with the r
     \begin{equation*}
       \prd{z,w:A}{q:x= z}{r:z= w} \big(\refl{x}\ct(q\ct r)= (\refl{x}\ct q)\ct r\big).
     \end{equation*}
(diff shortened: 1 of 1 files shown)


basics.tex
```diff
@@ -466,7 +466,7 @@ \section{Types are higher groupoids}
     \begin{equation*}
       \prd{z,w:A}{q:x= z}{r:z= w} \big(\refl{x}\ct(q\ct r)= (\refl{x}\ct q)\ct r\big).
     \end{equation*}
-    To construct an element of this type, let $D_2:\prd{x,z:A}{q:x=z} \type$ be the type family
+    To construct an element of this type, let $D_2:\prd{x,z:A} (x=z) \to \type$ be the type family
```
andrejbauer: `(x = z)` or `x = z`?

mikeshulman: I prefer (x=z); otherwise my eye takes a moment to reject the parsing x=(z\to\type).

andrejbauer: Eyes are precious, let's merge.

PR #795, opened 2015-06-02
Introduce and use \cong for isomorphisms of sets, close #793

### LINKED ISSUE #793: Theorem 10.2.10
In the statement and proof of theorem 10.2.10, I think the formula A \cong B should be A \simeq B.  (The proof refers to an "isomorphism" but A and B are sets; when this theorem is used in cor. 10.2.11, the proof refers to an "isomorphism" A \simeq B obtained using the theorem.)

mikeshulman:


mikeshulman:
bump

andrejbauer:
I can merge this, except for the minor detail that in `basics.tex` we promise to use `\cong` for isomorphisms of sets, and then use it for isomorphism of categories. Another point is that nobody will remember that `\cong` was introduced in `basics.tex` by the time they see its first (real) occurrence in Chapter 10. Should we just move the notation to Chapter 10 and make it local there?

mikeshulman:
Well, we don't promise in chapter 2 that `\cong` will _only_ mean isomorphisms of sets.  I like having it mentioned in chapter 2, but maybe we could mention there that it won't be used very much and recall it in chapter 10 before we use it?

andrejbauer:
Nah, either we leave it as is, or we move it all to chapter 10. Also, the way it is now we need to use the word "set" in chapter 2.

mikeshulman:
Sets are already mentioned in 3 other places in chapter 2: on source lines 153, 2277, and 2317.

mikeshulman:
(I guess this is another argument for moving parts of chapter 3 into chapter 1 in the second edition.)

andrejbauer:
Ok, I am convinced.

PR #800, opened 2015-06-03
Remove potential race conditions in travis push procedure

JasonGross:


andrejbauer:
I think there's little point in reviewing the Travis stuff you changed. Go ahead and just push whatever you're doing – it's not like anyone can detect problems without actually testing the changes with Travis, which you're doing already.

mikeshulman:
But I'm curious what led you to notice this?

JasonGross:
I was reading the `.travis.yml` file and noticed someone had set `UPDATE_ERRATA="yes"` in two of them; I had never intended it to be set on multiple travis runs for the same commit.  It's not really a big issue, but I figured I'd fix it.

PR #801, opened 2015-06-03
Add a script to auto-update nightly builds

### LINKED ISSUE #799: Configure travis to automatically post nightly builds
See discussion at #794.

JasonGross:
This closes #799.


JasonGross:
Should I just merge this, or does someone want to look it over?

mikeshulman:
Looks great to me, thanks!!  The only thing I don't understand is the "only make if there are no errata to be updated", can you explain that?

JasonGross:
If, for whatever reason, making the errata file fails, or if we decide to split up the travis job that marks errata and the travis job that pushes the pdfs, we only want to regenerate pdfs after the errata have been updated, or else we get an errata pdf with a pre-marked entry.  (It's not _that_ big of a deal if we don't; they'll be replaced anyway after travis pushes the errata commit.)

PR #802, opened 2015-06-04
Mentioning 'make errata.pdf'

danielgerigk:
It would be nice if you could instruct 'make' to include in the generated 'errata.pdf' only those errata which are more recent than some given version of the book.


mikeshulman:
There is a script in the repository called `filter-errata` which does exactly that (although I haven't tested it for some time, so it might have broken).

PR #803, opened 2015-06-07
A.2.5 Sigma-COMP rule remove unecessary primes

heatherclarke:


mikeshulman:
This looks good to me, but I'd like confirmation from someone more familiar with the appendix.  Why were those primes there to begin with?

cangiuli:
I don't remember, but almost certainly g's free variables were once named a and b, in which case it makes sense to call the pair (a',b'), so that the RHS is g[a',b'/a,b]. The pull request looks fine to me.

PR #806, opened 2015-06-12
for all pairs of equal elements

danielgerigk:


preliminaries.tex
```diff
@@ -1664,7 +1664,7 @@ \subsection{Path induction}
 
 Given a proof $p : \id{a}{b}$,
 path induction requires us to replace \emph{both} $a$ and $b$ with the same unknown element $x$; thus in order to define an element of a family
-$C$, for all pairs of elements of $A$, it suffices to define it on the diagonal.
+$C$, for all pairs of equal elements of $A$, it suffices to define it on the diagonal.
```
andrejbauer: This is a nice correction. I wonder if we need to be even more careful and explain that "for all pairs of equal elements" means "triples (a,b,p) where p : a = b"?

danielgerigk:
I would vote for "no, we do not need to be even more careful".

preliminaries.tex
danielgerigk: I would vote for "no, we do not need to be even more careful".

mikeshulman: It's a good question, but I think since this paragraph is just an _informal_ introduction to based path induction, we don't need to be more careful.

PR #807, opened 2015-06-15
replace "tuples" by "pairs"

danielgerigk:


danielgerigk:
I have no idea why the build fails. I didn't do anything evil, did I?

mikeshulman:
No, it failed to download latexmk for some reason.  I've kicked it to retry.

mikeshulman:
Ok, now it builds.

I'm actually not 100% sure about this change.  It's true that "tuple" is usually used only when the arity is large and it sounds somewhat weird to use it when n=2, but plain "pair" sounds maybe a little too informal or less precise to me.  Maybe "ordered pairs" would be better?  Other opinions?

danielgerigk:
Yes, that's true somehow.
In the second occurrence, people can deduce from context that what is meant by "pairs" are the terms of the form $(a,b)$ (the "canonical elements"), but indeed, in the first occurrence someone could object that in his understanding every term of type $A \times B$ is a pair.

danielgerigk:
I would also like to propose to alter the picture just before Lemma 2.3.4 in the following way: in the upper circle, I would like "f(x)" to be changed to "(x,f(x))", "f(y)" to be changed to "(y,f(y))", and the caption "f(p)" above the path from f(x) to f(y) to be deleted. I'm not very sure about the last point ( - removing the caption "f(p)" - ), but I'm "unhappy" with the fact that, as it stands, when you specialize to the non-dependent case f:A->B ( - i.e. when P is constant - ), you don't get precisely the f(p) defined in Lemma 2.2.1.

awodey:
I thunk "pair" is fine, but "ordered pair" is ok too.  not tuple, though.

awodey:
the picture should not be changed. the indexing elements are indicated by the position of the elements in the fibers above their respective indices.

basics.tex
```diff
@@ -540,7 +540,7 @@ \section{Types are higher groupoids}
 thus we have an operation $\Omega A\times \Omega A\to \Omega A$.
 More generally, the higher groupoid structure of $A$ gives $\Omega A$ the analogous structure of a ``higher group''.
 
-It can also be useful to consider the loop space\index{loop space!iterated}\index{iterated loop space} \emph{of} the loop space of $A$, which is the space of 2-dimensional loops on the identity loop at $a$.
+It can also be useful to consider the loop space\index{loop space!iterated}\index{iterated loop space} of the loop space of $A$, which is the space of 2-dimensional loops on the identity loop at $a$.
```
mikeshulman: This emphasis was intentional and should be kept.

mikeshulman:
I'm with @awodey, I think the picture is fine.  The label f(p) doesn't mean anything precise; at this point we're just talking intuitively about what sort of thing we _ought_ to be able to do.

Also, please don't add new unrelated changes to a pull request after it's been submitted; open a new pull request for those.

danielgerigk:
Ok, you've convinced me about the picture.
But I'll leave open the question of "pair" / "ordered pair", as I don't want to decide it.

@mikeshulman  Yes, I see. I'll initiate new pull requests, and close this one.

PR #808, opened 2015-06-17
fix typo

danielgerigk:


mikeshulman:
This is not a typo.

danielgerigk:
Okay, I'll resubmit that and call it a "minor correction"?

mikeshulman:
No, what I meant was, it's correct as written.

danielgerigk:
Okay... I don't quite get the reason for that, but it's not so important now either way.

mikeshulman:
It's just like the calculation which this one is "similar" to, two displays previous: some of the steps are judgmental but others are not, because e.g. the left and right unit and associativity laws are only propositional.

danielgerigk:
Yes, sure, but it's only the first equality sign that I changed; the latter one is still there and is still propositional.

mikeshulman:
Oh.  Hmm, my bad I guess.  So what you had would technically be correct.  But I think it would be misleading to write it that way, as it might give the impression (as it did to me!) that the whole equality is judgmental.  If we want to make that symbol a judgmental equality, then I think we need to put the second `=` on another line, so it stands out to the reader as being different.  Another possibility would be to omit the intermediate (refl . b) . (a . refl) entirely, since we're already omitting the rest of the computation.  Thoughts from anyone else?

danielgerigk:
Yes, omitting sounds good!

mikeshulman:
Feel free to make another PR for that, then.

danielgerigk:
Okay...

PR #810, opened 2015-06-18
if and only if

danielgerigk:
"Read logically, this says that two pairs are equal if and only if they are equal componentwise."


awodey:
it could as well say "just if" or "just in case".

awodey:
please provide descriptions of the pull requests.

danielgerigk:
@awodey Okay, I'll resubmit this one. Shall I write "just if", "just in case" or "if and only if"?

awodey:
let’s go with “just if”
thanks!

PR #812, opened 2015-06-18
propositional computation rules

danielgerigk:
I'm not sure if this looks nice enough... but I don't know how to do it better.

(I made the correction because both computation rules depend on both $p$ and $q$.)


mikeshulman:
Looks okay to me.

PR #814, opened 2015-06-18
\alpha \star' \beta = \beta \dot \alpha

danielgerigk:
remove intermediate step in equation

### DIFF
--- a/basics.tex
+++ b/basics.tex
@@ -618,7 +618,7 @@ On the other hand, we can define another horizontal composition analogously by
 \]
 and we similarly learn that
 \[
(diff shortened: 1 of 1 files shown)


mikeshulman:
Looks good to me, but does anyone else have an opinion?  (The point, carried over from a previous PR, is that we're already omitting most of the intermediate steps in the calculation, so there's no good reason to show just this one step, and it's confusing because this particular step happens to be a judgmental equality, yet if we indicate it as such it might give the mistaken impression that the whole equality is judgmental because it's all on one line.)

andrejbauer:
My opinion is that if we are showing any intermediate steps they should be the hard ones.

mikeshulman:
I'll take that as agreement, since the step we're showing right now is not a hard one.

PR #817, opened 2015-06-18
minor correction concerning "idtoeqv" as an elimination rule

danielgerigk:
$transport^{X \mapsto X}(p)$ would be of type "A -> B", not of type "A \equiv B".


danielgerigk:
Ah, that also applies to the propositional computation rule and the propositional uniqueness principle right below. But

PR #822, opened 2015-06-19
correction of a nasty equation concerning the higher groupoid structu…

danielgerigk:
…re of product types

### DIFF
--- a/basics.tex
+++ b/basics.tex
@@ -1322,16 +1322,16 @@ Or, written differently:
 \end{alignat*}
 All of these equations can be derived by using path induction on the given paths and then returning reflexivity.
 The same is true for the rest of the higher groupoid structure considered in \cref{sec:equality}, although it begins to get tedious to insert enough other coherence paths to yield an equation that will typecheck.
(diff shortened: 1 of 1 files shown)


danielgerigk:
I've triple-checked everything, and I'm still failing to see the book version as being correct.

mikeshulman:
I checked it in Coq at the time that I wrote it, because it was so complicated and I wanted to be sure to get it right: it's [here](https://github.com/HoTT/HoTT/blob/master/theories/Types/Prod.v#L172).  The Coq version uses `ap011` instead of `ap` because its `path_prod` is curried, but you can check that the book's version is also right by replacing the line

```coq
 = ap011 (path_prod u w) (concat_pp_p p q r) (concat_pp_p p' q' r')
```

by

```coq
  = ap (fun z => path_prod u w (fst z) (snd z)) (path_prod' (concat_pp_p p q r) (concat_pp_p p' q' r'))
```

danielgerigk:
Ok I trust you (and Coq). But the other corrections are correct, are they?

mikeshulman:
yes.  thanks!

mikeshulman:
If you want an explanation, `\pairpath(\ctassoc(p,q,r),\ctassoc(p',q',r'))` is a path between pairs of paths, whereas we need a path between paths between pairs.

danielgerigk:
thanks, now I feel dizzy.

PR #825, opened 2015-06-29
Fix the nightly build script (hopefully)

JasonGross:
Previously, the cherry-pick failed because rebasing did not discard the commit.
It's not entirely clear if we should use rebase or checkout+cherry-pick; they
should be equivalent, so it shouldn't matter.  I think checkout+cherry-pick
will be slightly more robust, e.g., with respect to other people modifying
history.

### DIFF
--- a/etc/ci/update_nightlies.sh
+++ b/etc/ci/update_nightlies.sh
@@ -106,8 +106,8 @@ fi
 
 git reset --hard || exit 1
 
-(git rebase origin gh-pages && git cherry-pick "$NIGHTLY_COMMIT" && "$DIR"/push_remote.sh HEAD:gh-pages) || exit 1
+(git remote update && git checkout origin/gh-pages && git cherry-pick "$NIGHTLY_COMMIT" && "$DIR"/push_remote.sh HEAD:gh-pages) || exit 1
 
-(cd book.wiki && git push origin HEAD:gh-pages) || exit 1
+(cd book.wiki && git push origin HEAD:master) || exit 1
 
 popd 1>/dev/null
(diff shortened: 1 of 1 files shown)


mikeshulman:
I don't quite understand what these commands are doing, can you explain (or add comments)?

JasonGross:
The `git remote` and `git diff` commands display more information about the state of the repo that I was curious about, and didn't have, when trying to debug this in my head.  (`--no-pager` means, roughly, "don't require keyboard input to display everything"; things like `less` and `more` are considered "pagers".)

```bash
(git remote update && git checkout origin/gh-pages && git cherry-pick "$NIGHTLY_COMMIT" && "$DIR"/push_remote.sh HEAD:gh-pages) || exit 1
```
This says "fetch any commits that have been pushed remotely between when we started this travis job, and now (`git remote update`), then checkout the origin `gh-pages` branch (which is where we want to push our nightly commit to), so that we can append (`git cherry-pick`) the commit with our nightly build files to that branch, and then push our `gh-pages`-with-nightly-commit to the origin (doing some work to account for the difference between `git@` and `https://`, and the auth token we're using).

Does that answer your wonderings?  Should some of this be added as a comment?  (Which bits?)

mikeshulman:
Why do we cherry-pick?  I mean, why are the nightly builds already committed to some other branch?

JasonGross:
Maybe it is worth being simpler, at the cost of being less robust in situations we don't expect to see.  I first `commit` and then `checkout`, because it's easier to make `checkout` robust in the face of conflicts with temporary files when you don't care about any of those temporary files, rather than when you care about some of them.  The way to stop caring about the generated files is to commit them somewhere, so that git keeps track of them for you.

mikeshulman:
Ah, I see.  Makes sense.

PR #829, opened 2015-07-02
fix typo

danielgerigk:


pthariensflame:
The `true`s and `false`s were correct before.

danielgerigk:
@pthariensflame i don't think so. ;)

mikeshulman:
I think @danielgerigk is right.  esucc originally went from odd to even, so its indexed version should go from paritynat(1) to paritynat(0).

pthariensflame:
Oh, OK.  I guess I was reading it wrong; sorry!

PR #830, opened 2015-07-02
stuff concerning "algebra 2-cells (from $f$ to $g$)"

danielgerigk:


mikeshulman:
Why the change to the diagram?

danielgerigk:
(I changed the "D"s to "C"s because "f,g : W -> C".)
Concerning the order of the two diagrams, I have no idea about the usual convention, but to me it somehow looked strange to have $s_f$ appear in the right diagram and $s_g$ in the left one.

PR #833, opened 2015-07-15
Fix the travis push script

JasonGross:
The push_remote script did not exist when checked out to another branch.
So save the current commit hash before checking out the other branch,
and go back to it before pushing.  Also move the relevant logic into a
separate file.

### DIFF
--- a/etc/ci/checkout-and-cherry-pick-and-push.sh
+++ b/etc/ci/checkout-and-cherry-pick-and-push.sh
@@ -1,0 +1,29 @@
+#!/bin/bash
+
+PS4='$ '
+set -x
+
+# in case we're run from out of git repo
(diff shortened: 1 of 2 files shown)


JasonGross:
Success!  The [nightly builds page](https://github.com/HoTT/book/wiki/Nightly-Builds) now links to the autogenerated pdfs!  Perhaps the [main wiki](https://github.com/HoTT/book/wiki) should no longer say "This wiki is not in use"?

andrejbauer:
Excellent, we should now fix the official book page to point there. I would still be in favor of linking directly to the nightly builds from the official book page, but for that we'd need fixed URLs.

mikeshulman:
I can't find where we discussed this before.  What did we decide to change the official book page to say?  Are we going to keep the "released" PDF versions separate from the nightly builds?

JasonGross:
The discussion is at #799.

PR #834, opened 2015-07-16
Clarify abuse of notation for equivalences, close #784

### LINKED ISSUE #784: Coercion from equivalences to functions
In some places we treat equivalences as functions, implicitly coercing away the witness of isequiv.  One such place is in the second and third sentences of the proof of Lemma 4.1.1:

"... we may assume that (f,e) is of the form idtoeqv(p) for some p:A=B. Then ... we may assume p is refl_A, in which case idtoeqv(p) is id_A".

Should we allow ourselves to do this?  (I think so.)  If so, ought we to be more explicit about it somewhere?  Where and how?

mikeshulman:


mikeshulman:
bump

PR #835, opened 2015-07-18
Use the new docker-based Travis infrastructure

JasonGross:
This allows us to cache the download of packages from ctan, which was a
step that was occasionally failing.  Now we will fail less, I think.

Followed the instructions at
http://docs.travis-ci.com/user/migrating-from-legacy/.

Also, it shaves off about 2 minutes from the initial configuration time.

I had to use an older version of TeXLive to do this, and had to manually download pgf/tikz and xcolor.


andrejbauer:
When I tell my mathematician friends that we install all of LaTeX every night they thought we were crazy. Little do they know that in not-so-distant a future the same thing will happen to them every time they fire up LaTeX.

spitters:
Will try to follow up later. Meanwhile, is the ubuntu+docker initiative
relevant to this?

spitters:
It's called snappy, but I don't see anyone combining it with Travis yet, so
let's wait for that.
https://developer.ubuntu.com/en/snappy/

PR #839, opened 2015-07-22
fix typo & variable renaming

danielgerigk:
I'm assuming that '\wtype' ("W-type") has about the same precedence as '\tprd' ("for all") and '\sm' ("exists").


mikeshulman:
Why the variable renaming?

danielgerigk:
I thought that "u:B(x) \rightarrow \wtype{x:A}B(x)" might be confusing, because of the "redefinition" of "x" as a local variable within "\wtype", overloading the previous meaning of "x". But now I think perhaps that's not a bug, but a feature...
But if we stick with "x" there, we could also change the "a" in line 414 to "x".

PR #840, opened 2015-07-22
solutions to exercises: 8.1 8.3 8.4 8.5

LuisScoccola:


exercise_solutions.tex
```diff
@@ -1441,6 +1441,245 @@ \subsection*{Solution to \cref{ex:acconn}}
...
+If we prove the equivalence without applying $\trunc 0{\blank}$ to both sides
+we are done.  We use induction on the natural numbers, with base case $1$.
+
+Suppose we have two pointed types $(A,a)$ and $(B,b)$.
+For the base case, we have $\eqv{\Omega(A \times B)}{\Omega(A) \times \Omega(B)}$
+because it is, by definition, the equivalence of pointed types:
+
+\[\eqv{ \Parens{(a,b)=(a,b), \refl{(a,b)}} }{\Parens{(a=a \times b=b), (\refl{a}, \refl{b})}}.\]
+
+By Univalence Axiom this is the same as proving the equality.
```
mikeshulman: I don't see where univalence comes in here?  I think this equivalence should be true without any need for univalence.  Maybe the point is that when we say "equivalence of pointed types" we mean by definition an equivalence of underlying types that respects the basepoints?

exercise_solutions.tex
```diff
@@ -1441,6 +1441,245 @@ \subsection*{Solution to \cref{ex:acconn}}
...
+Now we must give a $\susp \Sn^{n-1}$-indexed family of equalities between
+$\north_{n+1}$ and $\south_{n+1}$. By inductive hypothesis we have
+$i_{n-1} : \Sn^{n-1} \to \Sn^n$, so we can use:
+\[ \merid_{n+1} \circ i_{n-1} \]
+It is also interesting to note that this construction works in a more general
+setting: any function $f : A \to B$ induces a function $\susp f : \susp A \to \susp B$.
+
+Now is a good time to make a drawing to convince oneself that this is the
+right way to define the inclusions between consecutive spheres, and that this
+is the diagram intended in he exercise.
```
mikeshulman: Typo: "the exercise"

exercise_solutions.tex
```diff
@@ -1441,6 +1441,245 @@ \subsection*{Solution to \cref{ex:acconn}}
...
+
+Lets define the type $\Sn^{\infty}$ as the colimit of the diagram we just
+constructed. The constructors are:
+\begin{align*}
+  j_{\blank} &: \prd{n : \nat} \Sn^n \to \Sn^{\infty}\\
+  \glue_{\blank} &: \prd{n : \nat} \prd{x: \Sn^n} j_n(x) = j_{n+1}\circ i_n (x)
```
mikeshulman: I guess by `j_{n+1}\circ i_n (x)` you mean `(j_{n+1}\circ i_n)(x)`, but that's one of those things that's not immediately obvious how to parse.  I would tend to write `j_{n+1}(i_n (x))` instead.

exercise_solutions.tex
```diff
@@ -1441,6 +1441,245 @@ \subsection*{Solution to \cref{ex:acconn}}
...
+\[
+  \ind{\Sn^{\infty}} : \prd{C:\Sn^\infty \to \UU} \prd{J_{\blank} : \prd{n : \nat} \prd{x : \Sn^n} C(j_n(x))}
+  \Parens{\prd{n:\nat} \prd{x:\Sn^n} \dpath C {\glue_n(x)} {J_n(x)} {J_{n+1}\circ i_n (x)}}
+  \to \prd{e:\Sn^{\infty}} C(e)
+\]
+
+We also have the usual computational rules.
+Now, we can interpret the proof that we are going to give as follows.
+Each sphere is the ``equator'' of the next one. We say ``equator'' because it passes
+trough $\north$ and $\south$ so technically you have to rotate the bigger sphere for
```
mikeshulman: Typo: "through"

exercise_solutions.tex
```diff
@@ -1441,6 +1441,245 @@ \subsection*{Solution to \cref{ex:acconn}}
...
+the inclusion to be exactly in the equator.
```
mikeshulman: If you don't want to have to rotate, I guess you could say "meridian" instead of "equator".  (-:

exercise_solutions.tex
```diff
@@ -1441,6 +1441,245 @@ \subsection*{Solution to \cref{ex:acconn}}
...
+\node (N3) at (1.1,1.1) {};
+\node (N4) at (1.5,1.5) {};
+\draw[->] (N0) -- node[above]{\footnotesize $i_n$} (N1);
+\draw[->] (N0) -- node[left]{\footnotesize $j_n$} (N2);
+\draw[->] (N1) -- node[right]{\footnotesize $j_{n+1}$} (N2);
+\draw[double, double equal sign distance] (N3) -- node[left,above]{\footnotesize $glue_n$} (N4);
+\end{tikzpicture}
+\end{center}
+we get a path $x \defeq j_n(y) = j_{n+1} \circ i_n(y)$.
+But, as we will show, the image of $\Sn^n$ under $i_n$ is contractible,
```
mikeshulman: "the image of $\Sn^n$ under $i_n$ is contractible" is a little imprecise, which I suppose is okay since this is only an outline, but I think it might confuse the reader.  What you mean is that the "inclusion" `i_n` is nullhomotopic, right?

exercise_solutions.tex
```diff
@@ -1441,6 +1441,245 @@ \subsection*{Solution to \cref{ex:acconn}}
...
+
+To construct a $D_{\blank} : \prd{n:\nat} j_n(\north_n) = j_0(\north_0)$ we
+proceed by induction on $n$. For the base case we can 
+use $\refl{j_0(\north_0)}$. For the inductive case, we have by inductive
+hypothesis $j_n(\north_n) = j_0(\north_0)$. By our definition of $i_{\blank}$,
+we have that $\north_{n+1} \equiv i_n(\north_n)$. So $j_{n+1} (\north_{n+1})$
+equals $j_{n+1}\circ i_n(\north_n)$. By concatenation with $\glue_n(\north_n)$,
+we reduce our goal to the inductive hypothesis.
+
+Lets now show that the inclusion of $\Sn^n$ in $\Sn^{n+1}$ can be
```
mikeshulman: Typo: "Let's"

exercise_solutions.tex
```diff
@@ -1441,6 +1441,245 @@ \subsection*{Solution to \cref{ex:acconn}}
...
+continuously retracted to $\north_{n+1}$.
+That is, lets construct a homotopy:
```
mikeshulman: Again: "let's"

exercise_solutions.tex
```diff
@@ -1441,6 +1441,245 @@ \subsection*{Solution to \cref{ex:acconn}}
...
+Then we will be able to define the colimit as
+a higher inductive type.
+
+So, by induction on the natural numbers, we define $\Sn^{\blank} : \nat \to \UU$
+with the base case being the two point type $\bool$, and the inductive case
+being the iterated application of the suspension.
+
+Now we define the inclusions $i_n : \Sn^n \to \Sn^{n+1}$.
+For the base case we have to give a function $i_0 : \bool \to \susp \bool$.
+This is easy: send one point to $\north$ and the other to $\south$.
```
mikeshulman: I know the exercise says to start with `S^0`, but I notice that if we start with `S^{-1}` instead, which is empty, then the base case is even easier.  (-:

exercise_solutions.tex
```diff
@@ -1441,6 +1441,245 @@ \subsection*{Solution to \cref{ex:acconn}}
...
+By \cref{thm:transport-path} (and some straightforward computation) this reduces to:
+\[ i_n(\merid_n(x)) = \merid_{n+1}(\north_{n}) \]
+But, by our definition of $i_{\blank}$, and the computation rule of the
+suspension induction $i_n(\merid_n(x))$ equals $\merid_{n+1} \circ i_{n-1}(x)$.
+And, by inductive hypothesis, $i_{n-1}(x) = \north_n$, which gives us the
+desired result.
+
+Composing the two proofs we just gave we get a function:
+\[ J_{\blank} \defeq \glue_n(x)\ct \apfunc{j_{n+1}}{H_n(x)}\ct D_{n+1}
+    : \prd{n:\nat}\prd{x:\Sn^{\infty}} j_n(x) = j_0(\north_0). \]
```
mikeshulman: Typo: `x:\Sn^n`.

exercise_solutions.tex
```diff
@@ -1441,6 +1441,245 @@ \subsection*{Solution to \cref{ex:acconn}}
...
+\[ \prd{x:\Sn^{n-1}} \dpath {x\mapsto (i_n(x) = \north_{n+1})} {\merid_n(x)} {\refl{\north_{n+1}}} {(\merid_{n+1}(\north_n))^{-1}} \]
+By \cref{thm:transport-path} (and some straightforward computation) this reduces to:
+\[ i_n(\merid_n(x)) = \merid_{n+1}(\north_{n}) \]
+But, by our definition of $i_{\blank}$, and the computation rule of the
+suspension induction $i_n(\merid_n(x))$ equals $\merid_{n+1} \circ i_{n-1}(x)$.
+And, by inductive hypothesis, $i_{n-1}(x) = \north_n$, which gives us the
+desired result.
+
+Composing the two proofs we just gave we get a function:
+\[ J_{\blank} \defeq \glue_n(x)\ct \apfunc{j_{n+1}}{H_n(x)}\ct D_{n+1}
```
mikeshulman: I think if J is to have the type given that is a double dependent product, then it needs to be abstracted over n and x, i.e. to start with `\lam{n}{x}`.  In that case I would also write just `J \defeq ` at the beginning.  Or you could instead write `J_n(x) \defeq` and give the term you've written, but then the type should omit the two `\prd`s.

exercise_solutions.tex
mikeshulman: Also, just a comment (not mandatory to change): the macro `\prd` is smart so that you don't need to write it twice: you can just say `\prd{n:\nat}{x:\Sn^n}`.

exercise_solutions.tex
```diff
@@ -1441,6 +1441,245 @@ \subsection*{Solution to \cref{ex:acconn}}
...
+  {\glue_n(x)} {J_n(x)} {J_{n+1}\circ i_n (x)} \]
+By definition this is:
+\[ \prd{n:\nat} \prd{x:\Sn^n}
+    \transfib{x\mapsto (x = j_0(\north_0))}{\glue_n(x)}{J_n(x)}
+    = {J_{n+1}\circ i_n (x)} \]
+
+The LHS is equal to $\glue_{n}(x)^{-1}\ct J_n(x)$, and, by definition of
+$J_n(x)$, is just:
+\[ \glue_n(x)^{-1}\ct\glue_n(x)\ct\apfunc{j_{n+1}}{H_n(x)}\ct D_{n+1} =
+  \apfunc{j_{n+1}}{H_n(x)}\ct D_{n+1} \]
```
mikeshulman: This confused me at first; when you said "is just" and then displayed an equality, I thought you were saying that the previous displayed equation reduced to this equation.  But instead you are saying that the LHS of the previous displayed equation reduces to the LHS of this equation, which we know to be equal to the RHS of this equation by cancellation.  Maybe you could clarify the wording.

exercise_solutions.tex
```diff
@@ -1441,6 +1441,245 @@ \subsection*{Solution to \cref{ex:acconn}}
...
+  &= \glue_{n+1}(i_n(x))\ct\apfunc{j_{n+2}}{H_{n+1}(i_n(x))}\ct D_{n+2}\\
+  &= \glue_{n+1}(i_n(x))\ct\apfunc{j_{n+2}}{H_{n+1}(i_n(x))}\ct
+                                  \glue_{n+1}(\north_{n+1})^{-1}\ct D_{n+1}.
+\end{align*}
+
+So it suffices to show:
+\[ \apfunc{j_{n+1}}{H_n(x)} = \glue_{n+1}(i_n(x))\ct\apfunc{j_{n+2}}{H_{n+1}(i_n(x))}
+    \ct\glue_{n+1}(\north_{n+1}) \]
+
+But we remeber that we have the homotopy
```
mikeshulman: Typo: "remember"

exercise_solutions.tex
```diff
@@ -1441,6 +1441,245 @@ \subsection*{Solution to \cref{ex:acconn}}
...
+$\glue_{n+1} : j_{n+1} = j_{n+2}\circ i_{n+1}$, so, by a simple application
+of \cref{lem:htpy-natural}, the functoriality of $\apfunc{}{}$, and some standard
+cancelations, we reduce the goal to:
```
mikeshulman: Typo: "cancellations"

exercise_solutions.tex
```diff
@@ -1441,6 +1441,245 @@ \subsection*{Solution to \cref{ex:acconn}}
...
+\[ \apfunc{i_{n+1}}{H_n(x)} = H_{n+1}(i_n(x)) \]
```
mikeshulman: I would appreciate having this reduction spelled out in a little more detail.

exercise_solutions.tex
```diff
@@ -1441,6 +1441,245 @@ \subsection*{Solution to \cref{ex:acconn}}
...
+This can be done easily by induction in $\Sn^n$ using the definition of $H_{\blank}$.
+
+
+\subsection*{Solution to \cref{ex:contr-infinity-sphere-susp}}
+
+First we write down the type of the induction principle explicitly:
+
+\[ \ind{\Sn^\infty} : \prd{C:\Sn^\infty \to \UU} \prd{n:C(\north)}\prd{s:C(\south)}
```
mikeshulman: Again, you can leave out the intermediate `\prd`s (and doing so causes the book-style parentheses to get inserted).

exercise_solutions.tex
```diff
@@ -1441,6 +1441,245 @@ \subsection*{Solution to \cref{ex:acconn}}
...
+
+By \cref{cor:transport-path-prepost} it suffices to show
+$\refl{\north}\ct \merid(x) = \merid(\north)$. Canceling $\refl{\north}$ and
+applying $\merid$ to $p$ gets us the desired result.
+
+\subsection*{Solution to \cref{ex:unique-fiber}}
+
+We know that every two points $y_1,y_2 : Y$ are merely equal, because $Y$ is
+is connected. That is, we have a function
+$c : \prd{y_1,y_2:Y}\trunc {} {y_1 = y_2}$.
```
mikeshulman: This is by Exercise 7.6, I guess; is that fact stated anywhere else in the book?

exercise_solutions.tex
```diff
@@ -1441,6 +1441,245 @@ \subsection*{Solution to \cref{ex:acconn}}
...
+
+We note that it suffices to show that for any $y_1,y_2:Y$ we have
+$\trunc {} {\hfib{f}{y_1} = \hfib{f}{y_2}}$ because by Univalence
+$\hfib{f}{y_1} = \hfib{f}{y_2}$ is equivalent to $\hfib{f}{y_1}\simeq \hfib{f}{y_2}$
+and thus, by Univalence, they are equal.
```
mikeshulman: Again, the application of univalence is unnecessary here.

mikeshulman:
This is very nice!  Modulo my comments, it looks good to me, but since the writeup is somewhat lengthy I would appreciate it if someone else had the time to look it over as well.

One general comment: since `\blank` is shorthand for a lambda-abstraction, I guess `f_{\blank}` is shorthand for `\lam{x} f_x`, which by the uniqueness rule is equal to `f`.  So while it's not wrong to write `f_{\blank} \defeq`, it's simpler to just write `f \defeq`, and probably less confusing (especially if the type of `f` is actually a function of multiple variables).

exercise_solutions.tex
LuisScoccola: I thought we had to prove the equality, not only the equivalence. Fixed.

exercise_solutions.tex
LuisScoccola: That's true. In fact, by starting at 0 one has to repeat part of the argument. I prefer to start at 0 only because I like the family S^- to be defined over the natural numbers. But I can change it if it is not elegant to repeat part of an argument.

exercise_solutions.tex
LuisScoccola: I was refering to Remark 3.11.2. It is true that this is not the standard definition, so I added a comment in the solution.

exercise_solutions.tex
LuisScoccola: You are absolutely right. On the other hand, I noticed that (without using univalence) we can prove the equality instead of the equivalence. It is in fact simpler.

LuisScoccola:
I know that putting ```\blank``` can sometimes be confusing. I used it in this case because every function in the proof is indexed over the natural numbers, and ```\blank``` shows where the index ```n``` will be placed. When the function takes multiple arguments, the natural ```n``` is always the first one, so I think it is clear, but I can certainly be wrong.

exercise_solutions.tex
andrejbauer: You can also change the exercise in the book to have it start with 0 :-)

mikeshulman: I think it is fine as-is, I was just making an observation.

exercise_solutions.tex
```diff
@@ -1441,6 +1441,256 @@ \subsection*{Solution to \cref{ex:acconn}}
...
+$\refl{\north}\ct \merid(x) = \merid(\north)$. Canceling $\refl{\north}$ and
+applying $\merid$ to $p$ gets us the desired result.
+
+\subsection*{Solution to \cref{ex:unique-fiber}}
+
+We know that every two points $y_1,y_2 : Y$ are merely equal, because $Y$ is
+is connected. That is, we have a function
+$c : \prd{y_1,y_2:Y}\trunc {} {y_1 = y_2}$. This is by definition, if we use
+the definition stated in \textit{Remark $3.11.2$}. Otherwise we can use 
+\cref{ex:connectivity-inductively}.
```
mikeshulman: Ah!  I think the intended definition was that of \cref{ex:connectivity-inductively}, which is the standard meaning of "connencted" (= 0-connected).  The other is actually something that I think now should be changed; will open a new PR.

mikeshulman: Could you please update this reference, now that my PR has been merged?  I think the correct thing to point to is "the remark after \cref{thm:connected-pointed}" or else just \cref{ex:connectivity-inductively}.

exercise_solutions.tex
```diff
@@ -1441,6 +1441,256 @@ \subsection*{Solution to \cref{ex:acconn}}
...
+By \cref{cor:transport-path-prepost} it suffices to show
+$\refl{\north}\ct \merid(x) = \merid(\north)$. Canceling $\refl{\north}$ and
+applying $\merid$ to $p$ gets us the desired result.
+
+\subsection*{Solution to \cref{ex:unique-fiber}}
+
+We know that every two points $y_1,y_2 : Y$ are merely equal, because $Y$ is
+is connected. That is, we have a function
+$c : \prd{y_1,y_2:Y}\trunc {} {y_1 = y_2}$. This is by definition, if we use
+the definition stated in \textit{Remark $3.11.2$}. Otherwise we can use
```
mikeshulman: I think the citation of 3.11.2 should be removed in this case, but just for future reference, you shouldn't explicitly write in the number of any theorem or remark.  If it doesn't have a label yet in the source that you can use `\cref` on, just add such a label as part of the PR.  (You'll need to re-run `make main.labels` after adding such a label before it will show up in `exercise_solutions`.)

mikeshulman:
Thanks for your patience with all my suggestions.  It doesn't look like anyone else is going to look at this, so I'll go ahead and merge it.  Thanks for a nice write-up!

nicolaikraus:
I think these solutions are very good. There is one small typo that should be fixed. Sorry, I know I'm too late, so I'll create a quick pull request for that fix (I hope that's the way to do this? - I don't have too much experience with all the possibilities git offers).

PR #848, opened 2015-08-13
Fixed small typo in exercise solution

nicolaikraus:
typo fix in solution of pull request #840


mikeshulman:
That was the right thing to do.  Thanks!

PR #859, opened 2015-08-29
slight "simplification" in the definition of the cumulative hierarchy V

danielgerigk:
To me the definition feels slightly more pleasant if $A:U$ is assumed to be a set early on. But I think in the end this is only a matter of taste.


mikeshulman:
I like it better as it is.

danielgerigk:
Okay :)

PR #863, opened 2015-09-01
fix subscripts in Yoneda lemma, close #862

### LINKED ISSUE #862: yoneda lemma inverse proof functor misalignment?
In Theorem 9.5.4 the last part of the proof shows that the two natural transformation are inverses.
I am studying the first part of the proof and god stuck
```
αa′(f) = αa′(ya(a′,a)(f))
```
`f` is defined as of type `homA(a′, a)`.
The yoneda embedding is defined as `y: A → SetAop`
from this and `a: A` we deduce that `ya: SetAop` ie `ya` is a functor `Aop → Set`

From definition 9.2.1(ii) `F(a,b): homA(a,b) → homB(Fa, Fb)`
so in this context

`ya(a',a): homAop(a',a) → homSet(ya(a'),ya(a))` given that `ya` is a functor `Aop → Set`

but we have defined `f: homA(a',a)` and not `f: homAop(a',a)` which is what would be needed, so we can't apply f as it is done in the book with `ya(a',a)(f)`

mikeshulman:


mikeshulman:
Ok, I fixed the other subscript.  Anyone okay to merge this?

PR #871, opened 2015-11-06
Fix minor typos in "HoTT in Coq" introduction

neumanrq:


andrejbauer:
Hmm, this file is by now fairly obsolete. I wonder if anyone finds it useful.

neumanrq:
Oh, ok. I'm a beginner with Coq and found this resource very helpful. Probably other beginners may find it useful, too. Of course there are plenty of good Coq introductions available, but I like this one is directly related to HoTT. Maybe it could be kept at another place, though.

mikeshulman:
Well, in any case we may as well fix the misspellings.

PR #878, opened 2015-12-03
emphasize that UA is required for pi_1(S^1)=Z, close #736

### LINKED ISSUE #736: Is univalence really needed to prove pi_1(S^1)=Z?
Of course it is.

But the book doesn't seem to be explicit about that. (Is that a missed opportunity?)

This issue came up in our reading group meeting today, with questions by  some members of the reading group about that. They thought it would be possible to get away without univalance, by proving that pi_1(S^1) has the universal property of Z, using the techniques of the book.

This was my way to explain to them that univalence is unavoidable: Lemma 6.4.1 officially says that loop /= refl. But if you look at the proof, what it really proves is that  loop = refl iff every type is a set. 

Now "every type is a set" is indeed consistent with MLTT (this is Streicher's K axiom). 

Also, if we are to have  pi_1(S^1)=Z, then we cannot have loop = refl.

So, if we are to prove pi_1(S^1)=Z, then the axiom K must fail.

But the axiom K is independent. 

And univalence falsifies it. Thus, in any case, if we are to have pi_1(S^1)=Z, we cannot have K, and this is why univalence plays a fundamental role (because without it or something like it, K potentially holds).

I appreciate that this argument may be subtle, but on the other hand clever people wondered, and tried, to get rid of univalence in the argument given in the book (necessarily in vain).

Martin

mikeshulman:
Since the {rmk} is added at the end of a section, it doesn't change the numbering of anything else, so it is allowable as an addition to the first edition.

### DIFF
--- a/homotopy.tex
+++ b/homotopy.tex
@@ -785,6 +785,12 @@ This suggests a third approach.
 Of course, this proof also contains essentially the same elements as the previous two.
 Roughly, we can say that it unifies the proofs of \cref{thm:pi1s1-decode,lem:s1-encode-decode}, performing the requisite inductive argument only once in a generic case.
 
+\begin{rmk}
+  Note that all of the above proofs that $\pi_1(\Sn^1)=\Z$ use the univalence axiom in an essential way.
+  This is unavoidable: univalence or something like it is \emph{necessary} in order to prove $\pi_1(\Sn^1)=\Z$.
+  In the absence of univalence, it is consistent to assume the statement ``all types are sets'' (a.k.a.\ ``uniqueness of identity proofs'' or ``Axiom K'', as discussed in \cref{sec:hedberg}), and this statement implies instead that $\pi_1(\Sn^1) = \unit$.
+  In fact, the (non)triviality of $\pi_1(\Sn^1)$ detects exactly whether all types are sets: the proof of \cref{thm:loop-nontrivial} showed conversely that if $\lloop=\refl{\base}$ then all types are sets.
+\end{rmk}
 
 \section{Connectedness of suspensions}
 \label{sec:conn-susp}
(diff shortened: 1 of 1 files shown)


homotopy.tex
```diff
@@ -785,6 +785,12 @@ \subsection{The universal cover as an identity system}
 Of course, this proof also contains essentially the same elements as the previous two.
 Roughly, we can say that it unifies the proofs of \cref{thm:pi1s1-decode,lem:s1-encode-decode}, performing the requisite inductive argument only once in a generic case.
 
+\begin{rmk}
+  Note that all of the above proofs that $\pi_1(\Sn^1)=\Z$ use the univalence axiom in an essential way.
+  This is unavoidable: univalence or something like it is \emph{necessary} in order to prove $\pi_1(\Sn^1)=\Z$.
+  In the absence of univalence, it is consistent to assume the statement ``all types are sets'' (a.k.a.\ ``uniqueness of identity proofs'' or ``Axiom K'', as discussed in \cref{sec:hedberg}), and this statement implies instead that $\pi_1(\Sn^1) = \unit$.
+  In fact, the (non)triviality of $\pi_1(\Sn^1)$ detects exactly whether all types are sets: the proof of \cref{thm:loop-nontrivial} showed conversely that if $\lloop=\refl{\base}$ then all types are sets.
+\end{rmk}
```
EgbertRijke: The point in this remark seems to be that \pi_1(\Sn^1) is *equivalent* to \Z uses the univalence axiom in an essential way. The way it is currently formulated, someone might think that it is obvious that you need univalence to conclude \pi_1(\Sn^1)=\Z from the fact that they are equivalent, and such a person might fail to appreciate the point you are making. So I suggest using equivalences rather than equalities in this remark.

martinescardo: I didn't notice this, but now that you mention it, I agree with this remark, and I support the change.

mikeshulman: You're right. In fact, what I wrote is actually false: UIP doesn't imply
that pi1(S1) is *equal* to 1! I'll fix it.

mikeshulman: Fixed.

mikeshulman:
And with that, the version marker tops 1000 commits since the first edition!

PR #880, opened 2015-12-10
keep old nightly builds around for a day, close #846

### LINKED ISSUE #846: keep old nightly builds around for a day
As discussed at #799, since the saunders redirect only updates once a day.

mikeshulman:
Untested.  Is there any way to test a travis script aside from committing it and seeing what travis does?

### DIFF
--- a/etc/ci/update_nightlies.sh
+++ b/etc/ci/update_nightlies.sh
@@ -62,14 +62,15 @@ MASTER_COMMIT="$(git rev-parse HEAD)"
 
 git checkout -b gh-pages upstream/gh-pages || exit 1
 
-git rm -rf nightly || true
-rm -rf nightly || true
+OLD_NIGHTLIES="$(find nightly -type f -mtime +1)"
+git rm -rf $OLD_NIGHTLIES || true
+rm -rf $OLD_NIGHTLIES || true
 
-mkdir nightly
+mkdir -p nightly
 
 git add -f $PDFS || exit 1
 
-git mv $PDFS nightly/ || exit 1
+git mv -f $PDFS nightly/ || exit 1
 
 git commit -m "Update nightly builds (auto)" || exit 1
 NIGHTLY_COMMIT="$(git rev-parse HEAD)"
--- a/etc/ci/update_nightlies.sh
+++ b/etc/ci/update_nightlies.sh
@@ -62,11 +62,20 @@ MASTER_COMMIT="$(git rev-parse HEAD)"
 
 git checkout -b gh-pages upstream/gh-pages || exit 1
 
-OLD_NIGHTLIES="$(find nightly -type f -mtime +1)"
-git rm -rf $OLD_NIGHTLIES || true
-rm -rf $OLD_NIGHTLIES || true
-
 mkdir -p nightly
+
+# Delete files in /nightly not having a commit hash younger than 1 day
+# (The saunders redirects are only updated once a day.)
+YESTERDAY="$(date -d '1 day ago' +%s)"
+find nightly -type f | while read nfile
+do
+    HASH="$(echo $nfile | sed -r 's,^.*-([0-9a-z]+).pdf$,\1,')"
+    FILEDATE="$(git show -s --format=%ct $HASH 2>/dev/null)"
+    if [ -z $FILEDATE ] || [ $YESTERDAY -ge $FILEDATE ];
+    then
+	git rm -rf $nfile
+    fi
+done
 
 git add -f $PDFS || exit 1
 


etc/ci/update_nightlies.sh
```diff
@@ -62,14 +62,15 @@ MASTER_COMMIT="$(git rev-parse HEAD)"
 
 git checkout -b gh-pages upstream/gh-pages || exit 1
 
-git rm -rf nightly || true
-rm -rf nightly || true
+OLD_NIGHTLIES="$(find nightly -type f -mtime +1)"
```
JasonGross: You might want to add a note that the removal will fail if any of the file names contain spaces.  Also, add `2>/dev/null` to the `find` command, so we don't get spew to stderr if `nightly` doesn't exist.

mikeshulman: Oh ugh.  Should we do instead `find ... -exec git rm -rf {} + -exec rm -rf {} +`?

etc/ci/update_nightlies.sh
```diff
@@ -62,14 +62,15 @@ MASTER_COMMIT="$(git rev-parse HEAD)"
...
+git rm -rf $OLD_NIGHTLIES || true
```
JasonGross: It's slightly nicer to only run this command if `$OLD_NIGHTLIES` is nonempty; you can wrap it in `if [ ! -z "$OLD_NIGHTLIES" ] ...`.  (`git rm -rf` fails on an empty list; `rm -rf` does not.  No idea why.)

etc/ci/update_nightlies.sh
JasonGross: Something like that is probably better.

JasonGross:
Urg, `find -mtime` won't work; all of the files are freshly checked out.  You'll need to do something fancier, like parse the commit hash and check the commit date.

JasonGross:
Or you could do something like http://stackoverflow.com/a/2038768/377022 to update timestamps in nightly based on commit dates.

mikeshulman:
Why are we doing both `git rm` and `rm` anyway?  Doesn't `git rm` already remove files from the working directory as well?

etc/ci/update_nightlies.sh
JasonGross: You should probably also leave the `find` without the `exec` in there as well, so that travis will print out what files were found or not found, for debugging purposes.

JasonGross:
I don't think it's actually necessary to do both; I was too lazy to see if just doing `git rm -rf` was enough.

JasonGross:
> Is there any way to test a travis script aside from committing it and seeing what travis does?

You can turn travis on for your fork and push it there to see.  You can also run the commands in .travis.yml locally on your machine.

mikeshulman:
Ok, how about this?

etc/ci/update_nightlies.sh
```diff
@@ -62,14 +62,24 @@ MASTER_COMMIT="$(git rev-parse HEAD)"
...
+mkdir -p nightly
+
+# Delete files in /nightly not having a commit hash younger than 1 day
+# (The saunders redirects are only updated once a day.)
+YESTERDAY="$(date -d '1 day ago' +%s)"
+find nightly -type f | while read nfile
+do
+    HASH="$(echo $nfile | sed -r 's,^.*-([0-9a-z]+).pdf$,\1,')"
+    FILEDATE="$(git show -s --format=%ct $HASH 2>/dev/null)"
+    if [ -z $FILEDATE ] || [ $YESTERDAY -ge $FILEDATE ];
```
JasonGross: I would suggest quoting these variables (`$nfile`, `$HASH`, `$FILEDATE`, `$YESTERDAY`), to be on the safe side ("always quote every variable unless you have an explicit reason not to"), but it probably doesn't matter much.

JasonGross:
I'm a bit confused why `FILEDATE` is empty (https://travis-ci.org/HoTT/book/jobs/96147521#L8636), but the code seems reasonable to me.

peterlefanulumsdaine:
As far as I can see this seems good to merge, but I don’t have the Travis–fu to feel confident.  @JasonGross (or anyone else confident): am I right to read your comment above as approval to merge?

JasonGross:
Yes, my comment was approval, and the code still seems fine.  I am still confused about the date being empty, but if it turns out that this method doesn't distinguish the right files, it's still no worse than what we currently have.

JasonGross:
One possibility, which I'll note here for posterity, is that the clone on Travis, being a shallow clone, fetched not enough information to find the commit hash.  Even if this is the case, I don't think merging this hurts.

mikeshulman:
I guess in order to test whether this works, we have to make some other change to the book itself so that Travis will update the nightlies?

PR #883, opened 2016-01-07
add 'of' (as opposed to 'on')

jeffersoncarpenter:
This becomes clear from the context, but IMO the preposition should be made explicit


andrejbauer:
Actually, that's a rather poor description of what a polymorphic function is.

mikeshulman:
Well, let's merge the correction at least.  If anyone wants to write a better description that would be great.

PR #884, opened 2016-01-07
disambiguate 'id' and 'Id'

jeffersoncarpenter:
They differ only by capitalization.  A sentence like this one would have helped me as I read through the first chapter.


mikeshulman:
The lower-case one is used an order of magnitude more in the book (the identity type is usually written with "="), so I would be inclined to put the disambiguation remark near the introduction of the upper-case one instead.

andrejbauer:
I concur with @mikeshulman, let's have a remark next to `Id`.

jeffersoncarpenter:
Moved the remark, and updated the branch.

mikeshulman:
Thanks!

PR #886, opened 2016-01-24
Minor improvements

joliss:


categories.tex
```diff
@@ -17,7 +17,7 @@ \chapter{Category theory}
 If we add the requirement that the type $A_0$ of objects is a set, then we end up with a definition that behaves much like the traditional set-theoretic one.
 Following Toby Bartels, we call this notion a \emph{strict category}.
 \index{strict!category}%
-But we can also require a generalized version of the univalence axiom, identifying $(x=_{A_0} y)$ with the type $\mathsf{iso}(x,y)$ of isomorphisms from $x$ to $y$.
+But we can also instead require a generalized version of the univalence axiom, identifying $(x=_{A_0} y)$ with the type $\mathsf{iso}(x,y)$ of isomorphisms from $x$ to $y$.
```
andrejbauer: I am not a native speaker but it feels to me like there are too many words here. Maybe drop "also"?

equivalences.tex
```diff
@@ -717,7 +717,7 @@ \section{Closure properties of equivalences}
 \index{fibration}%
 \index{total!space}%
 Finally, we show that fiberwise equivalences can be characterized in terms of equivalences of total spaces.
-To explain the terminology, recall from \cref{sec:fibrations} that a type family $P:A\to\type$ can be viewed as a fibration over $A$ with total space $\sm{x:A} P(x)$, the fibration being is the projection $\proj1:\sm{x:A} P(x) \to A$.
+To explain the terminology, recall from \cref{sec:fibrations} that a type family $P:A\to\type$ can be viewed as a fibration over $A$ with total space $\sm{x:A} P(x)$, with the fibration being the projection $\proj1:\sm{x:A} P(x) \to A$.
```
andrejbauer: This introduces two "withs" one after the other. I do not like it.

categories.tex
joliss: We can drop the "also" if you want, though it feels like an acceptable turn of phrase to me ([Google seems to agree](https://www.google.com/search?q="we%20can%20also%20instead")). Let me know what you prefer.

equivalences.tex
joliss: Fixed.

categories.tex
mikeshulman: Why do you want to add "instead"?

andrejbauer: Without "instead" the text reads like we assume $A_0$ is a set *and* we also assume a generalized version of univalence. The word "instead" indicates an alternative. An alternative would be to say instead "Alternatively we can require ..."

mikeshulman: I think I like "Alternatively" best.  Maybe then we could also change the next sentence to say "Since we regard the latter choice as usually..."

joliss: Sounds good. I've updated this PR. (Or feel free to just commit your own version and close this. I don't have strong -- or useful -- opinions on the wording.)

PR #889, opened 2016-03-07
8.4.4(iii): mention that we need to equate these as pointed maps

fpvandoorn:
We need this, because otherwise we cannot prove that `f⁽⁶⁾` is identified with `Ω²(f)` (because to get a homotopy between `Ω(f⁽³⁾)` and `Ω²(f)` we need a *pointed* homotopy between `f⁽³⁾` and `Ω(f)`). Feel free to reword any of these changes; this might not be the best way to formulate it.

### DIFF
--- a/homotopy.tex
+++ b/homotopy.tex
@@ -1003,7 +1003,7 @@ space $Y$, the total space\index{total!space} $X$, and the fiber $F\defeq \hfib 
   \begin{enumerate}
   \item The fiber of $f^{(1)}\defeq \proj1 : \hfib f {y_0} \to X$ is equivalent to $\Omega Y$.\label{item:fibseq1}
   \item Similarly, the fiber of $f^{(2)} : \Omega Y \to \hfib f {y_0}$ is equivalent to $\Omega X$.\label{item:fibseq2}
-  \item Under these equivalences, the map $f^{(3)} : \Omega X\to \Omega Y$ is identified with $\Omega f\circ\rev{(\blank)}$.\label{item:fibseq3}
+  \item Under these equivalences, the pointed map $f^{(3)} : \Omega X\to \Omega Y$ is identified with the pointed map $\Omega f\circ\rev{(\blank)}$.\label{item:fibseq3}
   \end{enumerate}
 \end{lem}
 \begin{proof}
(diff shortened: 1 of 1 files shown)


homotopy.tex
```diff
@@ -1003,7 +1003,7 @@ \section{Fiber sequences and the long exact sequence}
   \begin{enumerate}
   \item The fiber of $f^{(1)}\defeq \proj1 : \hfib f {y_0} \to X$ is equivalent to $\Omega Y$.\label{item:fibseq1}
   \item Similarly, the fiber of $f^{(2)} : \Omega Y \to \hfib f {y_0}$ is equivalent to $\Omega X$.\label{item:fibseq2}
-  \item Under these equivalences, the map $f^{(3)} : \Omega X\to \Omega Y$ is identified with $\Omega f\circ\rev{(\blank)}$.\label{item:fibseq3}
+  \item Under these equivalences, the pointed map $f^{(3)} : \Omega X\to \Omega Y$ is identified with $\Omega f\circ\rev{(\blank)}$ (here we treat path inversion as pointed map of type $\Omega X\to \Omega X$).\label{item:fibseq3}
```
mikeshulman: Is the parenthetical really necessary?  (If we do keep it, I think it should be "as _a_ pointed map".)

homotopy.tex
```diff
@@ -1024,7 +1024,7 @@ \section{Fiber sequences and the long exact sequence}
   \cref{item:fibseq2} follows immediately by applying~\ref{item:fibseq1} to $f^{(1)}$ in place of $f$.
   % The resulting equivalence sends $(((x,p),p'),q') : \hfib{f^{(2)}}{(x_0,f_0)}$ to $\ap{f}{\opp {q'}} \ct p'$, while its inverse sends $s:x_0=x_0$ to $(((x_0,f_0), s), \refl{(x_0,f_0)})$.
   Since $(f^{(1)})_0 \defeq \refl{x_0}$, under this equivalence $f^{(3)}$ is identified with the map $\Omega X \to \hfib {f^{(1)}}{x_0}$ defined by $s \mapsto ((x_0,f_0),s)$.
-  Thus, when we compose with the previous equivalence $\hfib {f^{(1)}}{x_0} \eqvsym \Omega Y$, we see that $s$ maps to $\opp{f_0} \ct \ap{f}{\opp s} \ct f_0$, which is by definition $(\Omega f)(\opp s)$, giving~\ref{item:fibseq3}.
+  Thus, when we compose with the previous equivalence $\hfib {f^{(1)}}{x_0} \eqvsym \Omega Y$, we see that $s$ maps to $\opp{f_0} \ct \ap{f}{\opp s} \ct f_0$, which is by definition $(\Omega f)(\opp s)$. We omit the proof that the second component of these pointed maps are also identified.
```
mikeshulman: Should be "second components"

mikeshulman: But maybe it would be clearer to say something like "...the proof that this is an equality of pointed maps rather than just of functions"?

homotopy.tex
fpvandoorn: I don't know. I thought it might be useful, because we haven't introduced path inversion as a pointed map yet. But I'm okay with removing it.

homotopy.tex
fpvandoorn: I like your suggestion better than what I had. I'll change it once we decide what to do with the parenthetical.

homotopy.tex
fpvandoorn: Suggestion:
```
Under these equivalences, the pointed map $f^{(3)} : \Omega X\to \Omega Y$ 
is identified with the $\Omega f\circ\rev{(\blank)}$, treated as pointed map.
```

mikeshulman: How about

   Under these equivalences, the pointed map $f^{(3)} : \Omega X\to \Omega Y$ is identified with the pointed map $\Omega f\circ\rev{(\blank)}$.

fpvandoorn: Ok, I'll change it to that.

fpvandoorn:
Updated the commit.

mikeshulman:
Thanks!

PR #890, opened 2016-03-07
8.4.8: do k = 0 separately

fpvandoorn:
The proof didn't work for `k=0`.

### DIFF
--- a/homotopy.tex
+++ b/homotopy.tex
@@ -1153,12 +1153,12 @@ As an immediate application, we can now quantify in what way $n$-connectedness o
 \begin{cor}\label{thm:conn-pik}
   Let $f:A\to B$ be $n$-connected and $a:A$, and define $b\defeq f(a)$.  Then:
   \begin{enumerate}
-  \item If $k\le n$, then $\pi_k(f):\pi_k(A,a) \to \pi_k(B,b)$ is an isomorphism.
+  \item If $k\le n$, then $\pi_k(f):\pi_k(A,a) \to \pi_k(B,b)$ is an isomorphism.\label{item:conn-pik1}
   \item If $k=n+1$, then $\pi_k(f):\pi_k(A,a) \to \pi_k(B,b)$ is surjective.
   \end{enumerate}
 \end{cor}
 \begin{proof}
-  As part of the long exact sequence, for each $k$ we have an exact sequence
+  For $k=0$, part~\ref{item:conn-pik1} follows from Lemma~\ref{lem:connected-map-equiv-truncation}, noticing that $\pi_0(f)\equiv\trunc 0f$. For $k>0$ we have as part of the long exact sequence an exact sequence
   \[\xymatrix{\pi_k(\hfib f b) \ar[r]& \pi_k(A,a) \ar^f[r] & \pi_k(B,b) \ar[r] & \pi_{k-1}(\hfib f b).}\]
   Now since $f$ is $n$-connected, $\trunc n{\hfib f b}$ is contractible.
   Therefore, if $k\le n$, then $\pi_k(\hfib f b) = \trunc0{\Omega^k(\hfib f b)} = \Omega^k(\trunc k{\hfib f b})$ is also contractible.
--- a/homotopy.tex
+++ b/homotopy.tex
@@ -1158,8 +1158,8 @@ As an immediate application, we can now quantify in what way $n$-connectedness o
   \end{enumerate}
 \end{cor}
 \begin{proof}
-  For $k=0$, part~\ref{item:conn-pik1} follows from Lemma~\ref{lem:connected-map-equiv-truncation}, noticing that $\pi_0(f)\equiv\trunc 0f$.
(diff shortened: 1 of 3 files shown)


mikeshulman:
Because there is no \pi_{-1}, I guess.

I notice that the statement also doesn't say what values of n it applies to.  Does the proof work for n<0?

fpvandoorn:
Indeed.

Well, `k` has to be at least zero to make sense of `πₖ`, and `n ≥ k` in both parts.

fpvandoorn:
Never mind, that was a stupid comment. 

You're right, the proof of part (ii) also doesn't work for `n = -1` and `k = 0`. Shall I add a sentence for that as well? I didn't realize this, because in [my formalization](https://github.com/cmu-phil/Spectral/blob/bb5f719cd92b13e16eaf8f3569fbeac0e0977020/homotopy/LES_applications.hlean#L69-L70) I quantified for `n` only over the natural numbers.

mikeshulman:
Sure, why not.

Does the Lean HoTT library have a trunc-index type like we do in Coq?

fpvandoorn:
[It does](https://github.com/leanprover/lean/blob/master/hott/init/trunc.hlean#L18-L94). In the linked theorem statement the `n` is coerced to `ℕ₋₂`. I'll make the change later today.

fpvandoorn:
The book hasn't proven that if `f` is surjective, then `∥f∥ₙ` is surjective, right? This is very straightforward to prove completely, but it would require a paragraph or so. What shall I do? I could add it as an easy exercise to chapter 7.

mikeshulman:
Hmm, that might be the best choice.  I think it generalizes to say that if k≤n and f is k-connected, then so is ‖f‖ₙ --- right?

fpvandoorn:
Yes, that is a correct generalization (and Lean also agrees). Is it false without the condition that `k ≤ n`? 

I cannot prove it without that condition, but using Corollary 8.8.5 it doesn't seem too unreasonable. For `m ≤ n`, is `πₘ(∥f∥ₙ)` the same as `πₘ(f)` under the 'canonical' equivalence `πₘ(∥A∥ₙ) ≃ πₘ(A)`? Without the condition that `k ≤ n` we cannot have it as an exercise of chapter 7, though (unless we find a simpler proof).

mikeshulman:
Here is a proof using only chapter 7 technology: if `f` is `k`-connected, then by Lemma 7.5.14, `‖f‖_k` is an equivalence.  But if `n≤k`, then `‖f‖_n = ‖ ‖f‖_k ‖_n`; so it is also an equivalence, hence in particular `k`-connected.

fpvandoorn:
Ah, good point. So shall I make "if `f` is `k`-connected, then so is `‖f‖ₙ`" an exercise on chapter 7? Anything I need to do additionally when adding an exercise to the book? I presume that new exercises are always added at the bottom of the list?

mikeshulman:
Yes, new exercises are always added at the end, so as not to cause renumbering of the other exercises.  This does make the ordering a bit haphazard; in the second edition, we can think about the most logical ordering of the exercises.  If you feel like also adding a solution to the exercise in the `exercise-solutions` file, that's always appreciated.

fpvandoorn:
I've added the exercise and solution, used it in `8.4.8` and rebased the commits on top of `master`.

homotopy.tex
```diff
@@ -1153,12 +1153,14 @@ \section{Fiber sequences and the long exact sequence}
...
-  \item If $k\le n$, then $\pi_k(f):\pi_k(A,a) \to \pi_k(B,b)$ is an isomorphism.
-  \item If $k=n+1$, then $\pi_k(f):\pi_k(A,a) \to \pi_k(B,b)$ is surjective.
+  \item If $k\le n$, then $\pi_k(f):\pi_k(A,a) \to \pi_k(B,b)$ is an isomorphism.\label{item:conn-pik1}
+  \item If $k=n+1$, then $\pi_k(f):\pi_k(A,a) \to \pi_k(B,b)$ is surjective.\label{item:conn-pik2}
   \end{enumerate}
 \end{cor}
 \begin{proof}
-  As part of the long exact sequence, for each $k$ we have an exact sequence
+  For $k=0$, part~\ref{item:conn-pik1} follows from Lemma~\ref{lem:connected-map-equiv-truncation}, noticing that $\pi_0(f)\equiv\trunc 0f$.
+  For $k=0$ part~\ref{item:conn-pik2} follows from Exercise~\ref{ex:is-conn-trunc-functor}, noticing that a function is surjective iff it's $(-1)$-connected, by Lemma \ref{thm:minusoneconn-surjective}.
```
mikeshulman: Sorry, just noticed something else: we use cleveref, so instead of "Lemma \ref{...}" you can just write "\cref{...}" and the word "Lemma" will be added automatically.  Same for exercises.

fpvandoorn: Ah, that's nice. Then it makes more sense that some label names for lemmas start with `thm:`. I updated it.

mikeshulman:
Oh, oops!  I merged it but then a second after I merged it I realized that this deserves an erratum.  Can you make an erratum?  There are instructions in `CONTRIBUTING.md`.

PR #891, opened 2016-03-07
8.6.15: mention that we don't check respecting multiplication

fpvandoorn:


mikeshulman:
Is there much of anything to checking that?

fpvandoorn:
I am currently formalizing it in Lean, but I haven't finished it yet. Hence I don't know whether there are any bottlenecks, but I don't think so. I think every step can be shown to respect multiplication/concatenation quite straightforwardly, so I don't think it's necessary to give a detailed proof in the book.

mikeshulman:
Great, that would have been my expectation too.

PR #894, opened 2016-04-25
fix hole in the proof of Theorem 8.8.3

fpvandoorn:
In the last line of the proof of theorem 8.8.3, it says

> (...) In this case, we have \pi_k(\ap_f) = \pi_{k+1}(f), and it's given that \pi_{k+1}(f) is a bijection.

Indeed, it's given that `\pi_{k+1}(f) : \pi_{k+1}(A, a) -> \pi_{k+1}(B, f(a))` is an bijection. However, to be able to apply the induction hypothesis, we need to prove that for all `p : a = a` the map
`\pi_k(ap_f) : \pi_k(a = a, p) -> \pi_k(f(a) = f(a), \ap_f(p))`. The sentence in the book only proves this for `p ≡ refl_a`. To prove the full statement, we can generalize `p` to a path from `a` to an arbitrary point `p` and apply path induction. Then it follows from the givens.

Two remarks about my changes:
- I also changed the word "isomorphism" to bijection in the theorem statement. "isomorphism" sounds like there is more structure which the map preserves
- I also went through all of section 8.8 and distinguished between the maps `ap_f : a = b -> f(a) = f(b)` and `\Omega f : \Omega(A,a) -> \Omega(B, f(a))`. These maps are not definitionally the same (even when `b ≡ a`) since the latter map composes with reflexivity on both sides. This comes back in the proof of Theorem 8.8.3, because the two maps

```
\pi_k(\Omega f)  : \pi_k(x = x,p) \to \pi_k(f(x) = f(x),\Omega f(p))
\pi_k(\apfunc f) : \pi_k(x = x,p) \to \pi_k(f(x) = f(x),\apfunc f(p))
```

have different types, because the base points `\Omega f(p)` and `\apfunc f(p)` are not definitionally equal. This means we cannot even ask whether these maps are homotopic (however, they are homotopic if we compose either of the maps with a map which is a change of basepoint). 

(note: in the last sentence I write `\pi_k(\apfunc f) = \pi_k(\Omega f)`, where in this case the - implicit - base point is reflexivity. In this case, these functions have the same type, since `\Omega f(\refl_a)` and `\apfunc f(\refl_a)` are both definitionally `\refl_{f(a)}`. So in this case, this equality type-checks)

A Lean formalization of this proof is at the bottom of [this file](https://github.com/fpvandoorn/lean/blob/personal/hott/homotopy/homotopy_group.hlean).

### DIFF
--- a/errata.tex
+++ b/errata.tex
@@ -695,6 +695,13 @@ While the page numbering may differ between copies with different version marker
   & 474-g5289470
   & $\pi_3(\Sn^2)=\Z$ should be stated as \cref{thm:pi3s2}, following from \cref{cor:pis2-hopf,thm:pinsn}.\\
   %
+  \cref{thm:whiteheadn}
+  & % merge of 1026-g8c2bc9d
+  & There was a hole in the last line of the proof of \cref{thm:whiteheadn}. After applying the
+  induction hypothesis, it needs to be checked that for any $p : a = a$ the map $\pi_k(\apfunc
+  f):\pi_k(x = x,p) \to \pi_k(f(x) = f(x),\apfunc f(p))$ is a bijection, while the proof only
+  checked this for $p\equiv\refl{a}$.\\
+  %
   % Chapter 9
   %
   \cref{ct:functor}
--- a/homotopy.tex
+++ b/homotopy.tex
@@ -2426,7 +2426,7 @@ Now we can prove the truncated Whitehead's principle.
 \begin{thm}\label{thm:whiteheadn}
   Suppose $A$ and $B$ are $n$-types and $f:A\to B$ is such that
   \begin{enumerate}
-  \item $\trunc0f:\trunc0A \to \trunc0B$ is an isomorphism, and\label{item:wh0}
+  \item $\trunc0f:\trunc0A \to \trunc0B$ is a bijection, and\label{item:wh0}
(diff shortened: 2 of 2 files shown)


mikeshulman:
Thanks for this!  It may be a little while before I have a chance to look at it, but maybe someone else will...

JsonJayrock:
"there is more structure which the map preserves" ....or the code is preserving the structuring of the map 
~ BAX ~

peterlefanulumsdaine:
This looks excellent to me.  I’ve made [a couple of tweaks](https://github.com/peterlefanulumsdaine/book/commits/8.8.3): 

- changed wording of the erratum — generally the errata are written to be read with the pre-errata version of the text, i.e. like “X should be added”, not “X was missing”
- some style changes (prose and TeX) in the fixed proof

However, I can’t push my commit back to the pull request branch, due to permissions.  @fpvandoorn : could you pull my commits in (assuming you’re happy with them), and then I think this looks good to merge?  To pull mine into your branch, checkout your `8.8.3` branch, and then run:

    git remote add pll https://github.com/peterlefanulumsdaine/book.git
    git pull pll/8.8.3

then push back to your fork, and the pull request should update appropriately, I think?

andrejbauer:
I did it, no need for @fpvandoorn to do anything.

fpvandoorn:
Sorry. I forgot about this. Thanks Andrej and Peter for looking at my pull request.

PR #904, opened 2016-07-27
Fix math typo in cumulative hierarchy

abooij:
Having the `: V` outside the math environment typesets it as non-cursive, whereas the other occurrences of `V` are typeset cursively.


mikeshulman:
Looks good to me, but travis is choking, apparently because it can't find "nextpage.sty"?  Anyone with more travis-fu than I want to try to fix it?

mikeshulman:
I'm going to merge this change since it looks fine; travis's problem is something else.

PR #908, opened 2016-08-11
6.4: fix induction principle for S²

fpvandoorn:
There was a mistake/typo in the induction principle for `S²`, the type family of the dependent path `s` was wrong.

### DIFF
--- a/errata.tex
+++ b/errata.tex
@@ -537,6 +537,10 @@ While the page numbering may differ between copies with different version marker
   & 327-g7cbe31c
   & In the first sentence after the proof of \cref{thm:apd2}, ``$P:\Sn^2\to P$'' should be ``$P:\Sn^2\to\type$''.\\
   %
+  \cref{sec:circle}
+  & % merge of 1037-g9a1fc9d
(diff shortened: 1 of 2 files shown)


errata.tex
```diff
@@ -537,6 +537,10 @@
   & 327-g7cbe31c
   & In the first sentence after the proof of \cref{thm:apd2}, ``$P:\Sn^2\to P$'' should be ``$P:\Sn^2\to\type$''.\\
   %
+  \cref{sec:circle}
+  & % merge of 1037-g9a1fc9d
+  & In the sentence after the proof of \cref{thm:apd2}, the type family over which $s$ is a dependent path should be $\lam{p} \dpath P p b b$ instead of $P$.\\
```
mikeshulman: I would probably say "the type family *in* which", what do you think?

fpvandoorn: Good point. It's a path *over* `p` *in* the family `λp. (...)`. I'll change it.

fpvandoorn: Changed it and merged the commits.

PR #909, opened 2016-08-16
Obey style in Definition 4.2.1

koba-e964:
Type of \tau(a) is opposite.


mikeshulman:
Thanks.  I think this should have an erratum; can you add one?

koba-e964:
Sure. I added an erratum.

Issue #910, opened 2016-08-29
Ambiguous variable bindings in Exercise 10.13


koba-e964:
The formula
```
∀(x ∈ V).∀(y ∈ x).∃(z ∈ V).z ∈ y ⇒ ∃(c ∈ (∪x)^x).∀(y ∈ x).c(y) ∈ y
```
in Exercise 10.13 is ambiguous. I think the leftmost `∀(x ∈ V)` quantifies the entire formula, and the other quantifiers work as usual. Is it correct?
If it is correct, how about changing it to
```
∀(x ∈ V).[∀(y ∈ x).∃(z ∈ V).z ∈ y ⇒ ∃(c ∈ (∪x)^x).∀(y ∈ x).c(y) ∈ y]
```
(adding brackets)?

mikeshulman:
In chapter 1 we said that all binding operators like lambda, Pi, and Sigma scope over the entire expression unless delimited by parentheses.  We didn't explicitly mention this when introducing ∃ and ∀, but I would expect it to apply to them too.  So I would say it should be instead

    ∀(x ∈ V).(∀(y ∈ x).∃(z ∈ V).z ∈ y) ⇒ ∃(c ∈ (∪x)^x).∀(y ∈ x).c(y) ∈ y

koba-e964:
I see. I will make a pull request later. Does it need an erratum?

mikeshulman:
I would say yes, since as written it is actually incorrect under any possible convention.

It wouldn't be wrong to add extra parentheses around the whole expression for clarity:

    ∀(x ∈ V).((∀(y ∈ x).∃(z ∈ V).z ∈ y) ⇒ ∃(c ∈ (∪x)^x).∀(y ∈ x).c(y) ∈ y)

Maybe make the outer ones `\Big` to set them apart.  If you think that would help, feel free to do it too.

PR #911, opened 2016-08-30
Fix ambiguous variable bindings in Exercise 10.12, 10.13

koba-e964:
Fix https://github.com/HoTT/book/issues/910.

### DIFF
--- a/errata.tex
+++ b/errata.tex
@@ -769,6 +769,9 @@ While the page numbering may differ between copies with different version marker
   \cref{thm:VisCST}
   & 706-ged2c765
   & In the proof of the function set axiom, ``the types of elements $[u] \mono V$ and $[u] \mono V$'' should be ``the types of members $[u] \mono V$ and $[v] \mono V$.''\\
+  \cref{ex:strong-collection}
+  & % merge 0c4f868
+  & Extra parentheses are needed to make the formula unambiguous. \\
   \cref{ex:choice-cumulative-hierarchy-choice}
   & % merge c395527
   & Extra parentheses are needed to make the formula unambiguous. \\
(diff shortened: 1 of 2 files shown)


errata.tex
```diff
@@ -769,6 +769,9 @@
   \cref{thm:VisCST}
   & 706-ged2c765
   & In the proof of the function set axiom, ``the types of elements $[u] \mono V$ and $[u] \mono V$'' should be ``the types of members $[u] \mono V$ and $[v] \mono V$.''\\
+  \cref{ex:strong-collection}
+  & % merge 0c4f868
+  & Extra parentheses are needed to make the formula unambiguous. \\
```
mikeshulman: Might as well be explicit here about where the parentheses go?

koba-e964:
I see. I added some information.

errata.tex
```diff
@@ -769,6 +769,12 @@
   \cref{thm:VisCST}
   & 706-ged2c765
   & In the proof of the function set axiom, ``the types of elements $[u] \mono V$ and $[u] \mono V$'' should be ``the types of members $[u] \mono V$ and $[v] \mono V$.''\\
+  \cref{ex:strong-collection}
+  & % merge 0c4f868
+  & Extra parentheses around $\fall{x\in v}\exis{y} R(x,y)$ are needed to make the formula unambiguous. \\
+  \cref{ex:choice-cumulative-hierarchy-choice}
+  & % merge c395527
+  & Extra parentheses around $\fall{y\in x}\exis{z\in V} z\in y$ are needed to make the formula unambiguous. In addition, the big parentheses after $\fall{x\in V}$ are added for clarity.\\
```
mikeshulman: Sorry to be such a nitpicker, but I wouldn't mention the "for clarity" parentheses.  But if you disagree, feel free to say so.  My rule of thumb is to think about who is reading the errata: it's not people who are reading the correct version, it's people who have the incorrect version and need to know what is wrong with it.  Ideally, we should tell them exactly what's mathematically wrong and how to fix it (though space doesn't always allow us to be fully detailed); but I don't think they need to know about cosmetic changes that we also made in the updated version.

koba-e964:
I've thought that errata should enumerate the difference in changes, but now I understand and agree. I fixed it.

PR #922, opened 2016-09-29
Fix \choice{\infty,\infty} in Exercise 7.8

### LINKED ISSUE #915: \choice{\infty,\infty} in Exercise 7.8
`\choice{\infty,\infty}` is
```
\prd{X:\set}{Y:X\to\UU}
  \Parens{\prd{x:X} Y(x)}
  \to
  \prd{x:X} Y(x)
```
, which is obvious by `\idfunc[\prd{x:X} Y(x)]` and not Theorem 2.15.7.

koba-e964:
Fix #915.

### DIFF
--- a/errata.tex
+++ b/errata.tex
@@ -655,6 +655,10 @@ While the page numbering may differ between copies with different version marker
   & This exercise needs more precise definitions of ``diagram'' and ``colimit''.\\
   %
   \cref{ex:acnm}
+  & % merge of bb5f959
+  & $\choice{\infty,\infty}$ is not \cref{thm:ttac}, but the identity function.\\
+  %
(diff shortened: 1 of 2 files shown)


mikeshulman:
Thanks!  Now that I look at it again, I wonder whether it would be worth keeping some remark about 2.15.7.  I figure that whoever wrote that exercise was thinking about a versions of AC_{\infty,\infty} formulated analogously to (3.8.1) rather than (3.8.3), and in that case it would be 2.15.7.  Would it be worth remarking on that?

koba-e964:
I agree with you on the point that remarking would be useful. I'll try writing this.

mikeshulman:
Thanks!

PR #931, opened 2017-01-24
Fix typo in natural numbers addition example

Jaetan:
Section 1.9 on natural numbers gives the example of integer addition for primitive recursion with a function of several variables. The defining equation: \add(\suc(m),n) &\jdeq \suc(\add(m,n)) uses the "next step" function:

  c_s & : \nat \to (\nat \to \nat) \to (\nat \to \nat) \\
  c_s(m,g)(n) & \defeq \suc(g(n)).

The currying of g should be made explicit to match the type \nat \to \nat for the second parameter of c_s, e.g. c_s(m, g(m))(n) & \defeq \suc(g(m)(n)).

### DIFF
--- a/errata.tex
+++ b/errata.tex
@@ -229,6 +229,10 @@ While the page numbering may differ between copies with different version marker
   & 111-g1e868fa
   & When translating English to type theory, ``unnamed variables'' are unnamed in English but must be named in type theory.\\
   %
+  \cref{sec:inductive-types}
+  & % merge of bf74e1eb426ed3eed1f41368c992d2ac52e11e8b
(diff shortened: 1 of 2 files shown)


mikeshulman:
That doesn't make sense.  According to the type of `c_s`, its first argument `m` is of type `nat`, its second argument `g` is of type `nat -> nat`, and its third argument `n` is of type `nat`; then we compute its output of type `nat` by applying `g` to `n` and then taking the successor.  The left-hand side of a defining equation for a function has to be applied to plain variables such as `m`, `g`, and `n`; it doesn't make sense to write `g(m)` there.

Jaetan:
You are right, there is no error in the formula as written in the book. I got confused thinking that `g` had to be the addition of integers `N -> N -> N` and then noticing that `c_s` did not use its parameter `m`. I now realize my mistake.

PR #932, opened 2017-02-07
small grammar correction (between vs. among)

OliveiraHermogenes:
Although "between" can be used to compare more than 2 things in some specific cases, I don't think this sentence qualifies as one of them. Furthermore, when I first read the sentence, I kept wondering if it was meant a relationship between the $n$-types and some other thing (which should be inferred from context) or among the $n$-types themselves. I suppose that it was meant the latter.

### DIFF
--- a/logic.tex
+++ b/logic.tex
@@ -105,7 +105,7 @@ Similarly, the defining property of a $1$-type is that it has no non-trivial pat
 \end{defn}
 
 Similarly, we can define $2$-types, $3$-types, and so on.
-We will define the general notion of $n$-type inductively in \cref{cha:hlevels}, and study the relationships between them.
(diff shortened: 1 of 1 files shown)


mikeshulman:
Thanks for the suggestion.  Even less ambiguous might be "relationships between n-types for different values of n"; what do you think?

OliveiraHermogenes:
@mikeshulman Yes, that sounds even better. Thanks for your quick reply.

mikeshulman:
Do you want to update this PR or make a new one?

OliveiraHermogenes:
@mikeshulman I don't know. I am familiar with git but I just started using Github. I guess it would be easier if you just ignore the pull request and implement the changes youself, no? After all, it is just a very minor issue.

PR #934, opened 2017-03-31
sentence-revision

CL123123:
The description of the README.md is clear and easy to follow. But in the last paragraph, there are some confusion when I read it. I make some revision on the sentence.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,5 +52,5 @@ replaced with your name):
 Then use whatever tools you normally do to compile LaTeX. The main LaTeX files are called 
 `hott-XXX.tex`. But you really should have `make`, you know.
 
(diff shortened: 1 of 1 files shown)


mikeshulman:
I think "then" is rather less clear than "so that".

andrejbauer:
I was taught that "need" is a [modal verb](http://english.stackexchange.com/questions/29409/why-use-need-not-instead-of-do-not-need-to). I am not a native speaker, but I'd stick with the modal verb here.

mikeshulman:
To be sure, the original version is grammatically correct.  But since the modal verb might be confusing for other non-native speakers who haven't been taught about it, I don't see any harm in clarifying.  Actually, how about clarifying the first part of the sentence too in a different way: #936.

PR #936, opened 2017-04-01
readme grammar

mikeshulman:


andrejbauer:
Did you really mean to write `Once `make` has been run once`?

andrejbauer:
I am not sure that avoiding perfectly good grammatical constructs really helps anyone. I am fine with the verison we have now. I don't object to the PR either, but someone else should accept it if they think this is an improvement.

PR #944, opened 2017-09-18
Add notation clarification, close #943

### LINKED ISSUE #943: Confusing Notation in Section 1.6
On page 43,
hott-1 6-confusing-notation

It might not be obvious to some readers that f is being obtained by application of a function. That is, it is not completely clear that the lambda expression that takes x is being applied to the lambda expression that takes f.

mikeshulman:


andrejbauer:
I am against this change (I read #943). If the issue is that we are displaying a higher-order beta-redex, i.e., a lambda applied to a lambda, then that issue should be dealt with elsewhere, where we introduce lambda calculus. 

If we react like this everytime someone finds confusing a notation which is actually exact, then there will be no end to remarks and clarifications. Figuring out what the expression means is a learning experience.

To me this feels a bit like someone being surprised by a nested sum because they've only seen single sums. They should have been shown a nested sum when they were first taught sums.

andrejbauer:
Having read more carefully the surrounding text, I think there is in fact a small problem. The text claims that a function is substituted for `f`, but the problematic formula does not substitute anything, it *applies* a function to an argument. And it is in fact misleading to teach people that substitution is the same thing as an application. I would suggest that we just delete the displayed formula.

mikeshulman:
@andrejbauer Note that this *is* the same chapter in which we introduce lambda-calculus.  I'm not against saying something about higher-order beta-redexes in section 1.2 or 1.4 (if someone can come up with a natural-seeming example to put there), but even if we do, the reader may not yet have fully internalized the idea by the time they get to section 1.6 where this happens.  I think it's reasonable to do some more hand-holding throughout chapter 1.

You're right, though, that confusing substitution and application is a more serious issue!  I assume whoever wrote this did it because we haven't introduced any notation for substitution, and beta-reducing the displayed application does perform the intended substitution.  Recall, though, that in order to avoid talking about (and introducing a notation for) substitution, sections 1.4 and 1.6 define dependent function and dependent pair types using type families that are defined as non-dependent *functions* into the universe.  So officially, in the language of the book, the dependent pair type `\tsm{f:A\to B} \tprd{x:A} R(x,f(x))` in the codomain of `\ac` really does involve the *function* `\lamu{f:A\to B} \tprd{x:A} R(x,f(x))`.  So I don't think the displayed equation is terrible, as long as we gloss it correctly in the text.  How about

```
...the type $\prd{x:A} R(x,\fst(g(x)))$ is the result of applying
the type family $\lamu{f:A\to B} \tprd{x:A} R(x,f(x))$ being summed
over in the codomain of \ac to the function $\lamu{x:A} \fst(g(x))$.
```

?

andrejbauer:
Yes, saying "applying" is a lot better than "substituting", and it clarifies the problematic expression at the same time. Does @jtmayer have an opinion?

jtmayer:
Looks good.

mikeshulman:
Ok, PR updated.

PR #952, opened 2017-10-11
Fix introduction to nat, close #946

### LINKED ISSUE #946: Technical error in first sentence of section 1.9
The first sentence of section 1.9 states "The rules we have introduced so far do not allow us to construct any infinite types" which is technically speaking invalid. For instance the type \prod_{A:\mathcal{U}} (A -> A) -> (A -> A) can be constructed after section 1.4. Micheal Shulman pointed out to me that even the first universe has infinitely many inhabitants so there are even simpler examples. Perhaps this should be reworded to something like "The rules introduced so far do not allow us to construct any natural examples of infinite types" or something like that? I'm not sure what would best replace the sentence.

mikeshulman:


jkingdon:
Looks good to me. In addition to fixing an error, it also better motivates what we are doing and foreshadows a later part of the book ("foreshadowing: your clue to quality literature").

PR #959, opened 2017-10-30
Don't refer to p which is not defined

jkingdon:
In the section "2.2 Functions are functors", Lemma 2.2.1 does not contain any reference to `p` but the second proof refers to `p` without any definition. I think this change might be a way to clarify the situation, but please take a look as I'm not confident that I understand the notational conventions or what is or is not supposed to be implicit.

If this change is accepted, I'm not sure something this minor needs to be mentioned in the errata, but again I'm not sure where the dividing line is for that.

### DIFF
--- a/basics.tex
+++ b/basics.tex
@@ -712,7 +712,8 @@ The notation $\apfunc f$ can be read either as the \underline{ap}plication of $f
 \end{proof}
 
 \begin{proof}[Second proof]
-  By induction, it suffices to assume $p$ is $\refl{x}$.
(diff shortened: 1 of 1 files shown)


mikeshulman:
I think I would be inclined to be even more explicit, like "to define `\apfunc{f}(p)` for all `p:x=y`, it suffices to assume..."  What do you think?

I agree this doesn't need an erratum, since it's basically a clarification of wording rather than a correction of an error.

jkingdon:
More explicit makes sense when it comes to introducing a variable. I've revised the pull request according to your suggestion.

mikeshulman:
Thanks!

PR #962, opened 2017-12-18
Order of arguments in 2.1.2

streawkceur:
It seems that, when applying the two path inductions, one doesn't immediately obtain an order of arguments as written before, but instead one as I wrote in the patch.

### DIFF
--- a/basics.tex
+++ b/basics.tex
@@ -274,9 +274,9 @@ We will write out the next few proofs in both styles, to help the reader become 
 Next we prove the transitivity of equality, or equivalently we ``concatenate paths''.
 
 \begin{lem}\label{lem:concat}
-  For every type $A$ and every $x,y,z:A$ there is a function
+  For every type $A$ and every $x,y,z:A$, there is a function
   \begin{equation*}
-  (x= y) \to   (y= z)\to (x=  z)
+  (x= y) \to   (y= z)\to (x=  z),
   \end{equation*}
   written $p \mapsto q \mapsto p\ct q$, such that $\refl{x}\ct \refl{x}\jdeq \refl{x}$ for any $x:A$.
   We call $p\ct q$ the \define{concatenation} or \define{composite} of $p$ and $q$.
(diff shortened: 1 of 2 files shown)


andrejbauer:
If I understand you correctly, you worry that the theorem states

    ∏ (x y z : A) . x = y → y = z → x = z                          (1)

but the proof constructs

    ∏ (x y : A ., x = y → ∏ (z : A) . y = z → x = z              (2)

In that case, the problem starts right at the beginning of the proof, when the proof sets out to construct an element of type (2) without a warning. It might be better to warn the reader *at the beginning* of the proof that something is going on, rather than to attach an "unexplained" fix at the end.

mikeshulman:
I agree with @andrejbauer if a fix is needed.  But is a change really necessary?  Is this actually confusing, or just pendantic?

streawkceur:
I can confirm that my learning partner and me were pretty confused and somewhat frustrated by the first proof of 2.1.2 (it took us two hours to discuss it completely - in contrast to the second proof, which we agreed with very quickly), but I agree that the order of the arguments was probably only a minor problem. Note that, at the time, we only very recently had learned the definition of path induction (and had essentially no intuition for it), so checking the proof reduced to comparing the symbols. I felt that this was being made more complicated than necessary by moving quite far away from the formal description of path induction. It also didn't really become clear to us from the formulation (before we understood it completely) that there is actually no real "trick" involved, but it is a straightforward application of two path inductions (and, in fact, changing the order of the argument is in a sense the only part of the proof that doesn't follow forcedly by doing the unique thing possible with path inductions and refl!).

I have written an alternative version which sticks more closely to the definitions, which I hope avoids these problems and which seems to be more easily understandable to me and my learning partner.

mikeshulman:
Thanks!  But I think your proposed change is too big; let's just add a sentence at the beginning of the proof alerting the reader of the approach we'll take.  (Also, I don't think an erratum is necessary, since this isn't a mathematical error, just a clarification of exposition.)

mikeshulman:
I've made a simpler suggestion in #963.

martinescardo:
This simpler suggestion looks good to me.

PR #966, opened 2018-02-14
fix definition of double^w, close #965

### LINKED ISSUE #965: Wrong definition for the return type at paragraph 5.3
At paragraph 5.3, I think the definition for C while defining the function _double_ using W-types is wrong, a lambda is missing. Instead, it should be

\lambda a^{2}.
\Pi f^{B(a) -> Nw}.
\Pi g^{B(a) -> Nw}.
Nw

At least in my formalization, I have to. Otherwise, while defining e_1 as 

lambda f.
lambda g.
succW (succW (g (star)))

won't be well-typed since e_1 needs to compute B(1) so that g(star) is well-typed. Thus, I have the impression that using rec2 as defined in paragraph 1.8 cannot work here.

mikeshulman:


induction.tex
```diff
@@ -313,9 +313,9 @@ \section{\texorpdfstring{$\w$}{W}-types}
 which will represent the recurrence for the $\dbl$ function; for simplicity we denote the type family $\rec\bool(\bbU,\emptyt,\unit)$ by $B$.
 
 Clearly, $e$ will be a function taking $a : \bool$ as its first argument. The next step is to perform case analysis on $a$ and proceed based on whether it is $\bfalse$ or $\btrue$. This suggests the following form
-\[ e \defeq \lamu{a:\bool} \rec\bool(C,e_0,e_1,a) \]
+\[ e \defeq \lamu{a:\bool} \ind\bool(C,e_0,e_1,a) \]
 where
-\[ C \defeq \prd{f : B(a) \to \natw}{g : B(a) \to \natw} \natw. \]
+\[C \defeq \lamu{a:\bool} \prd{f : B(a) \to \natw}{g : B(a) \to \natw} \natw\]
```
mikeshulman, line 318:
```suggestion
\[C \defeq \lamu{a:\bool} \prd{f : B(a) \to \natw}{g : B(a) \to \natw} \natw.\]
```
Oops, the closing period got lost in the merge.

PR #968, opened 2018-02-28
labels for Examples 3.1.2 and 3.1.3

marcbezem:
See also HoTT/HoTT/#918 and HoTT/book/#967


mikeshulman:
I assume you're trying to deal with the fact that #967 isn't merged yet.  I think I'll go ahead and merge #967 myself (instead of this).  Even though general policy is that someone else should do it, all the other people with that privilege are kind of slow about responding to PRs, and in this case the change is so minor and its absence is breaking HoTT/Coq that I think it's justified.

marcbezem:
Thanks. This gets us out of the critical section and frees the lock on HoTTBook.v so that I can add some work to that file.

PR #971, opened 2018-03-04
Partial Spanish Translation of Reading_HoTT_in_Coq(Fixed)

dbucci24:
This is the first part of my contribution, then I will continue with the other parts.
I have fixed the problems of the previous PR.
 
-Translation from line 804 to 1075
-1179 words

### DIFF
--- a/coq_introduction/Reading_HoTT_in_Coq.v
+++ b/coq_introduction/Reading_HoTT_in_Coq.v
@@ -977,7 +977,7 @@ El tipo "Verdadero" tiene un constructor y el tipo "Falso" tiene cero.
 *)
 (** *** No operador *)
 (**
-En HoTT, el no operator indica que el elemento de un tipo puede ser
+En HoTT, el operador no indica que el elemento de un tipo puede ser
 mapeado a los elementos del tipo vacío (cero).
 *)
 
--- a/coq_introduction/Reading_HoTT_in_Coq.v
+++ b/coq_introduction/Reading_HoTT_in_Coq.v
@@ -1058,9 +1058,9 @@ Debido a que este archivo está usando Coq "estándar", no podemos demostrar
 tipos inductivos mayores.
 *)
 
-(** ** Propiedades de las rutas *)
+(** ** Propiedades de lod caminos *)
 (**
-Para cada elemento de un tipo, hay una ruta constante.  Esta es la misma
+Para cada elemento de un tipo, hay un camino constante.  Esta es la misma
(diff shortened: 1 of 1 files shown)


martinescardo:
Looks good now. However, wouldn't "camino" be better than "ruta" for path? I am not sure. Better to ask Spanish academics before committing to one particular translation. Maybe "ruta" is already established as the terminology in Spanish academic circles?

coq_introduction/Reading_HoTT_in_Coq.v
```diff
@@ -1058,9 +1058,9 @@ Debido a que este archivo está usando Coq "estándar", no podemos demostrar
 tipos inductivos mayores.
 *)
 
-(** ** Propiedades de las rutas *)
+(** ** Propiedades de lod caminos *)
```
martinescardo: lod -> los.

martinescardo:
https://es.wikipedia.org/wiki/Homotop%C3%ADa has "camino" in the caption for the top-right picture.

martinescardo:
@andrejbauer @mikeshulman As far as the accuracy of the translation goes, this can be merged, I think.

mikeshulman:
Sorry, I should have put my comment on the other PR here.

1. I think a fork or separate project is a more appropriate place for a translation.
2. Unless I'm misreading this PR, it would actually *change* the current master to be in Spanish instead of English, which is definitely not what we want.

dbucci24:
@mikeshulman So I do the translation on a fork, and when I complete it, what do I do?

dbucci24:
@mikeshulman who will merge my translations? myself?

mikeshulman:
If a translation were going to be kept in the main repository, there would have to be a separate directory structure for it.  However, I would rather leave it in a fork or separate project, with a link from appropriate places.  The reason is that different people are responsible for the translation (you and everyone else who contributes to it) than for the original text.  If someone reports a problem with the translation, it should go directly to the translators as maintainers of the translated version rather than having to pass through us, for convenience if nothing else.  And if the translation were in the main repository, there would be an implicit responsibility of the authors to keep it up to date and in sync, which I don't want to take on -- we have a hard enough time maintaining the original version.

dbucci24:
ok, this is the fork : https://github.com/dbucci24/book-1

I'll work there, is it ok?

mikeshulman:
Sounds good to me!  When you've got something stable and substantial, let us know and I expect we can add links to appropriate places.

dbucci24:
Great. I'm working on the translation on my fork. When I have finished it, I will let you know!

Regards!

PR #977, opened 2018-03-10
Fix typo in 6.5.3 lemma

WojciechKarpiel:
1. I added `\label` to fixed lemma so I can reference it in errata. It was the only lemma in HIT (6th) chapter without `\label`
2. Visually (in PDF) numberings did not change but `make labelcheck` prints `Some label numbers have changed since the first edition!` both before and after my change. Output of `make labelcheck` is identical before and after my change so I assume it's okay.


mikeshulman:
Actually that's not a typo; it means that g'(u) is defined to be b_0 *for all u*.  Just like when in algebra you define a function by writing f(x) = x^2.

It does seem that `make labelcheck` is currently broken; probably the format of aux files changed again.  d-:

WojciechKarpiel:
My bad, sorry

PR #981, opened 2018-04-13
exercise: nondependent funext implies dependent funext

mikeshulman:


mikeshulman:
Thanks; do you think we should credit him for this fact in the Notes of chapter 4 as well?  Or in the exercise?

martinescardo:
I think the notes.

mikeshulman:
Ok, added to notes.

PR #982, opened 2018-04-17
Fix missing `apfunc` in lemma `transport-s1-code`

WojciechKarpiel:
Hello!
In proof of `transport-s1-code` the `transport-compose` lemma was applied incorrectly and there was ordinary application of function instead of application to path which made proof incorrectly-typed
This time it's real typo fix :)

### DIFF
--- a/errata.tex
+++ b/errata.tex
@@ -667,6 +667,10 @@ While the page numbering may differ between copies with different version marker
   %
   % Chapter 8
   %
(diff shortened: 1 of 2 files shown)


mikeshulman:
Nope, this one is also intended.  (-:  See after Lemma 2.2.1: "We will often write ap_f(p) as simply f(p)."

PR #983, opened 2018-04-18
Add & remove periods after formulas

wdeweijer:
I believe https://github.com/HoTT/book/commit/48c7c7db90a3b1fa9ceff34a48d6872307f24fce fixes a small typo, but I might be wrong. All other changes are small typographical fixes.


martinescardo:
I think this belief is correct.

mikeshulman:
Oh, that typo fix should be errata'ed.  Do you want to do it?

PR #991, opened 2018-05-31
Fix typo, missing parenthesis in transport

jonaprieto:
`\trans{p}{_}` is a unary-function.

### DIFF
--- a/basics.tex
+++ b/basics.tex
@@ -1511,7 +1511,7 @@ This appears in the statement of action of transport on $\Sigma$-types, which is
   \end{equation*}
   For any path $p:x=y$ and any $(u,z):\sm{u:P(x)} Q(x,u)$ we have
   \begin{equation*}
-    \trans{p}{u,z}=\big(\trans{p}{u},\,\trans{\pairpath(p,\refl{\trans pu})}{z}\big).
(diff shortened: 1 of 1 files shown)


mikeshulman:
I think this is not a typo, but just a standard abuse of notation in mathematics; I'm sure we've used it in many other places in the book too.  I suppose we should remark on it somewhere though, if we didn't.

jonaprieto:
> I'm sure we've used it in many other places in the book too
This is the only usage of this transport with pairs. I searched in all the text files with a regex.

mikeshulman:
I'm not just talking about transport, but about arbitrary functions of two variables.

jonaprieto:
Oh yes, you're right. I found other places in the book with this same issue, e.g. some lemmas in Section 6.12, but for consistency, I'd use parenthesis.

mikeshulman:
Actually, I think one of the points of the book is that using type theory as a foundation for mathematics doesn't require us to be any more pedantic about such things than mathematicians usually are.  But, as I said, it might be helpful to mention this abuse of notation somewhere towards the beginning.

mikeshulman:
Actually, this is already mentioned in section 1.5:

> However, later on in the book we will often write $g(a,b)$ both for functions on a product and for curried functions of two variables.

PR #998, opened 2018-09-12
Fix typo

koba-e964:
Fixed a typo.
Does this fix need an erratum?


mikeshulman:
I'm going to say no erratum is needed for this, since I can't imagine anyone misinterpreting the uncorrected version.  But thanks for the fix.

Issue #999, opened 2018-09-13
A suggestion of an exercise about split monomorphisms


koba-e964:
I believe that the following are logically equivalent:
(a) LEM,
(b) for every sets A and B, if A is inhabited then for every monomorphism f: A -> B, f is also a split monomorphism. (i.e. f has a retraction.)
(a) -> (b) can be proved just like in classical set theory. (b) -> (a) can be proved by considering a monomorphism f: 1 + P -> 2 (`f(inl(*)) \defeq 0, f(inr(_)) \defeq 1`) for a mere proposition P. If we pick a retraction g of f, checking if `g(1) = inl(*)` decides whether P holds.
How about adding this as an exercise? If this is added, I think appending it to the last of Exercise 10 is good.

mikeshulman:
That sounds good to me.  Yes, new exercises should be appended to the list of exercises, and Chapter 10 seems a reasonable place (although Chapter 3 might also be).

PR #1000, opened 2018-09-14
Add an exercise about split monomorphisms

koba-e964:
It closes https://github.com/HoTT/book/issues/999.
The exercise is appended to the list of Exercise 10 and will be Exercise 10.18.


mikeshulman:
Thanks!

PR #1001, opened 2018-09-19
[ hits ] some fixes in lemmas in Section 6.12

jonaprieto:


hits.tex
```diff
@@ -1945,8 +1945,9 @@ \section{The flattening lemma}
     \transfib{x\mapsto Q}{\pp(b)}{c(f(b),y)}
     &= c(f(b),y) \tag{by \cref{thm:trans-trivial}}\\
     &= c(g(b),D(b)(y)) \tag{by $p(b,y)$}\\
-    &= c(g(b),\transfib{P}{\pp(b)}{y}). \tag{by \cref{thm:transport-is-given}}
+    &= c(g(b),\transfib{P}{\pp(b)}{y}) \tag{by $\opp{\apfunc{c(g(b),-)}(q)}$}
   \end{align}
+  where $q : \transfib{P}{\pp(b)}{y}) = D(b)(y) $ comes from \cref{thm:transport-is-given}.
```
mikeshulman: I don't think this change is necessary (and it's certainly not a typo).  "by" doesn't necessarily mean that the equality cited is precisely the one appearing here; as a Coq tactic it's more like `rewrite` than `apply`.

hits.tex
```diff
@@ -2008,9 +2009,9 @@ \section{The flattening lemma}
     \notag \\
     &= c(f(b),y)
     \tag{by \cref{thm:trans-trivial}}\\
-    &= c(f(b),D(b)(y))
+    &= c(g(b),D(b)(y))
    \tag{by $p(b,y)$}\\
-    &= c(f(b),\trans{\pp(b)}{y}).
+    &= c(g(b),\trans{\pp(b)}{y}).
```
mikeshulman: These typo fixes should have errata.

mikeshulman:
It looks like the fork/branch that this PR came from has been deleted.  @jonaprieto, are you able/willing to update it, or would you like someone else to submit a replacement PR?

jonaprieto:
> It looks like the fork/branch that this PR came from has been deleted. @jonaprieto, are you able/willing to update it, or would you like someone else to submit a replacement PR?

I'll give it a shot.

hits.tex
jonaprieto: > don't think this change is necessary (and it's certainly not a typo)

It's not a typo, it's a suggestion that facilitates a second read of the proof 
of the calculation, let me know if you want me to remove this. 

> the equality cited is precisely the one appearing here; as a Coq tactic it's more like `rewrite` than `apply`.

Sure, nevertheless I think this may help someone else like me for translating this calculation to say Agda.

jonaprieto:
I'm gonna close this PR, but I'll read your comments on the open conversion @mikeshulman in case you have.

hits.tex
mikeshulman: Personally, I think maximal detail isn't always maximally clear for a human reader.  In particular, I find it more confusing to introduce the new variable `q` and then define it later.  The fact that Agda requires you to be more explicit about how the cited lemma is used is a defect in Agda, not something we should imitate in informal type theory.  (-:

PR #1006, opened 2018-10-12
Fix a tiny typo

iblech:
(The intermediate value theorem is about continuous functions having a zero, while the mean value theorem is about differentiable functions.)

1. Shall I add an entry to the errata list?
2. Shall I change the label of the exercise (from `ex:mean-value-theorem` to `ex:intermediate-value-theorem`)?


mikeshulman:
Thanks!  I see that I fixed this in the exercise itself back in 08942310, but failed to notice this reference to it.  I don't think an erratum is really necessary.  I'd lean towards not changing the label, since it might break various things.

PR #1012, opened 2019-01-31
Misspell: indiscernability -> indiscernibility

Mehran-Baghi:
Hi

* Reference:
https://plato.stanford.edu/entries/identity-indiscernible/
3rd paragraph of "1. Formulating the Principle":
"The converse of the Principle, x=y → ∀F(Fx ↔ Fy), is called the Indiscernibility of Identicals."

* I "grep"ed all the files and fixed all of them.

* I guess this is a trivial change so I didn't edit the errata.


mikeshulman:
It seems that you are right about this one.  I'm very surprised that no one caught this before.  Didn't we run a spellcheck?

PR #1013, opened 2019-01-31
Remove the extra comma

Mehran-Baghi:
* Because higher is the adjective of universe.
* This is a trivial change so I didn't edit the errata.

### DIFF
--- a/preliminaries.tex
+++ b/preliminaries.tex
@@ -1507,7 +1507,7 @@ Note also that we can use the universes in type theory to represent ``higher ord
 For example, we can represent the proposition \emph{for all properties $P : A \to \UU$, if $P(a)$ then $P(b)$} as
 \[ \prd{P : A \to \UU} P(a) \to P(b) \]
 where $A : \UU$ and $a,b : A$.
-However, \emph{a priori} this proposition lives in a different, higher, universe than the
+However, \emph{a priori} this proposition lives in a different, higher universe than the
 propositions we are quantifying over; that is
 \[ \Parens{\prd{P : A \to \UU_i} P(a) \to P(b)} : \UU_{i+1}. \]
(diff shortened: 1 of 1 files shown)


Mehran-Baghi:
Yes both different and higher are adjectives of universe. that justifies the first comma (different, higher).
These are coordinate adjectives.
https://www.thoughtco.com/what-is-coordinate-adjectives-1689739

but since higher is the last adjective we shouldn't separate it from its noun by another comma.

An example:

"The air was thick and wet. A **warm, dense fog** had settled over the paddies and there was the stillness that precedes rain."
(Tim O'Brien, "The Things They Carried." Esquire, 1987)

mikeshulman:
I don't agree that they are coordinate adjectives: they are both adjectives of "universe", but they do not have equal status in describing it.  The word "higher" gives further emphasis to "different": the universe is not *just* different, but *furthermore* higher.  In particular, it would not make sense to reverse them: "lives in a higher, different universe" would not make sense since of course a higher universe must be different.  I'm sorry I don't know the technical grammatical term for this usage of a comma, but it's point 3 at https://owl.purdue.edu/owl/general_writing/punctuation/commas/extended_rules_for_commas.html: 

> Use a pair of commas in the middle of a sentence to set off clauses, phrases, and words that are not essential to the meaning of the sentence. Use one comma before to indicate the beginning of the pause and one at the end to indicate the end of the pause.

The sentence could be written more verbosely with "lives in a different, indeed a higher, universe", in which case both commas are clearly necessary.  Condensing the phrase "indeed a higher" to the single word "higher" doesn't change its grammatical role in the sentence.

Mehran-Baghi:
You are right.
Thank you for your time and patience.

mikeshulman:
Thank you in turn for listening, and for your interest in improving the book!

PR #1027, opened 2019-05-06
Fixed minor typo in introduction.tex

anqurvanillapy:
`s/entirely analogously/entirely analogous/g`


mikeshulman:
I think this is correct as written; we need an adverb there, not an adjective.

anqurvanillapy:
Well I think *analogous* is used as an adjective for *inductively defined spaces*, since when the adverbs (*entirely* and *analogously*) are deleted, the sentence is left to look weird due to the *to* word.  When deleting *analogous* used as an adjective, the *to* part should be removed along with it. For instance:

> Inductively defined spaces can be characterized, ~~analogous to classical examples~~

This looks weird:

> Inductively defined spaces can be characterized, ~~analogously~~ to classical examples

But actually this is not a great deal.  If you insist I will close this PR.

mikeshulman:
Well, yes, of course if you delete 'analogously' you have to delete the entire phrase that goes along with it, but that doesn't say anything about whether it's an adjective or an adverb.  What's being said is not that the inductively defined *spaces* (a noun) are analogous to classical examples thereof, but that the *characterization* (a verb) is analogous to the characterization of classical examples, and to modify a verb you need an adverb.

However, I see that it may not be clear from the phrasing that the analogy is to the *characterization* of classical examples rather than to the examples themselves.  Perhaps "to the cases of classical examples" or even "to the characterizations of classical examples" would be better?

anqurvanillapy:
Oops sorry for my misreading. But as you suggest, I would think a kinda naively extended clause may be nice:

> Inductively defined spaces can be characterized, which is entirely analogous to the cases of classical examples.

anqurvanillapy:
Thanks for your help here!  Closing it for my previous misreading and a unnecessity.

PR #1031, opened 2019-06-10
Add missing lim() in reals.tex

abooij:


andrejbauer:
LGTM

mikeshulman:
Looks good to me too, but should have an erratum.  @abooij , do you have time to add one?

abooij:
I added another commit that adds something to the errata, but I'm not sure it's correct, so please double-check!

PR #1036, opened 2019-08-24
Update canonical inhabitant symbol in Table 1 to match the Index of Symbols.

jsnwest:


mikeshulman:
No, that's not right.  `\ttt` is type-theoretic notation for the *inhabitant* of the unit type; this `*` is a common homotopy-theoretic notation for the unit type *itself*.

jsnwest:
Ok, thanks for the insight.

PR #1037, opened 2019-09-12
Add ebook-wide and ebook-narrow options

liyishuai:
Fix coherence with
https://github.com/HoTT/book/blob/ca5e8fac377d1bd1ae13f2e4687a4ab3da3ff785/opt-ebook.tex#L2

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@
 DEFAULTTOPTEX = hott-online.tex
 
 # Top-level LaTeX files from which HoTT book can be generated
-TOPTEXFILES = $(DEFAULTTOPTEX) hott-ustrade.tex hott-letter.tex hott-letter-exercises.tex hott-a4.tex hott-a4-exercises.tex hott-ebook.tex hott-ebook-wide.tex hott-arxiv.tex
+TOPTEXFILES = $(DEFAULTTOPTEX) hott-ustrade.tex hott-letter.tex hott-letter-exercises.tex hott-a4.tex hott-a4-exercises.tex hott-ebook.tex hott-ebook-wide.tex hott-ebook-narrow.tex hott-arxiv.tex
 
 # LaTeX files that actually comprise the book
 # (that is, all of them except configuration)
@@ -37,6 +37,7 @@ OPTFILES = opt-letter.tex \
 	   opt-ustrade.tex \
 	   opt-ebook.tex \
 	   opt-ebook-wide.tex \
+	   opt-ebook-narrow.tex \
 	   opt-color.tex \
 	   opt-black-white.tex \
 	   opt-cover.tex \
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@
 DEFAULTTOPTEX = hott-online.tex
 
 # Top-level LaTeX files from which HoTT book can be generated
-TOPTEXFILES = $(DEFAULTTOPTEX) hott-ustrade.tex hott-letter.tex hott-letter-exercises.tex hott-a4.tex hott-a4-exercises.tex hott-ebook.tex hott-arxiv.tex
+TOPTEXFILES = $(DEFAULTTOPTEX) hott-ustrade.tex hott-letter.tex hott-letter-exercises.tex hott-a4.tex hott-a4-exercises.tex hott-ebook.tex hott-ebook-wide.tex hott-arxiv.tex
 
 # LaTeX files that actually comprise the book
 # (that is, all of them except configuration)
@@ -36,6 +36,7 @@ OPTFILES = opt-letter.tex \
 	   opt-a4.tex \
 	   opt-ustrade.tex \
 	   opt-ebook.tex \
+	   opt-ebook-wide.tex \
 	   opt-color.tex \
 	   opt-black-white.tex \
 	   opt-cover.tex \
--- a/hott-ebook-narrow.tex
+++ b/hott-ebook-narrow.tex
@@ -1,0 +1,28 @@
+% -*- mode: LaTeX -*-
+% This is a main LaTeX file for generating the HoTT book
+% for ebook readers. It has low resolution images and small
+% margins. The links are black in this version.
+
+% **** OPTIONS ****
+
+% INCLUDE A COVER
+\input{opt-cover}
+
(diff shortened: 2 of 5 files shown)


andrejbauer:
Please, can you explain why you think this change is necessary? I'd be very apprehensive of changing the format of the page.

andrejbauer:
As far as I can tell, the formats in the various `optXYZ.tex` files are ok, but the comments accompanying them are wrong (for instance, the US trade size claims to be the letter format).

liyishuai:
The ratio of 4.8:7.9 is too narrow for Kindle, which is about 3:4.

liyishuai:
Ah, Kindle and Kindle Fire have different ratio. Is it worth adding another option instead?

liyishuai:
I've updated the proposed size to 4.8 by 7.2, which does not change the width of pages, and corresponds to US trade size. The current ratio is not US letter either.

andrejbauer:
I see, thank you for explaining. In that case, I would suggest that we do this diffrerently. For each common e-reader we can simply create the most suitable format. Call these by the name of the reader, e.g., `kindle-3`. If you agree, I would propose that we close this PR and you open another one that does it. We can still have the "default" ebook format.

We can even link to all of the various formats from the book web page. @mikeshulman, how does that sound to you?

andrejbauer:
I suppose what I am saying is that `ebook-wide` is not very descriptive, whereas something like `ebook-kindle-fire` is.

mikeshulman:
Well, there are a *lot* of ebook readers with a lot of different aspect ratios, and for any given aspect ratio there'll be at least several readers.  So if we name the formats after readers, we'll have potentially a *lot* of formats with a lot of redundancy.  If we want to be more fine-grained than just having two formats like `ebook-narrow` and `ebook-wide` (which I personally think ought to be sufficient, but whatever), I would suggest naming them by aspect ratio.  Then the book page can include a list of common readers having each ratio.

liyishuai:
Is there a way to parameterize the options if they're only different in ratio? We could have multiple build target based on same template.
Also, feel free to push modifications to my branch.

JasonGross:
> Is there a way to parameterize the options if they're only different in ratio? We could have multiple build target based on same template.

Here is a way to get the information into a `.tex` file from the filename:
```bash
$ cat Makefile
```
```Makefile
EBOOK_RATIOS := 8_6 8_4.69 16_9

EBOOK_PDFS := $(EBOOK_RATIOS:%=hott-ebook-%.pdf)
EBOOK_TEXS := $(EBOOK_PDFS:.pdf=.tex)

.PHONY: all

all: $(EBOOK_PDFS)

%.pdf: %.tex main.tex
	pdflatex $<

$(EBOOK_TEXS) : hott-ebook-%.tex : template.tex
	sed 's/@RATIO@/$*/' $< > $@
```
```bash
$ cat template.tex
```
```tex
\def\ebookratio{@RATIO@}
\input{main}
```
```bash
$ cat main.tex
```
```tex
\documentclass{article}

\def\firstunder#1_#2#{#1}
\def\secondunder#1_#2#{#2}
\edef\ebookwidth{\expandafter\firstunder\ebookratio{}}
\edef\ebookheight{\expandafter\secondunder\ebookratio{}}

\begin{document}

The ratio is \ebookwidth\space to \ebookheight.

\end{document}
```

andrejbauer:
While @JasonGross's solution is c00l and all, I dislike it. Too mach hackery.

A good solution might be to have several aspect ratios (one file for each, no Makefile and sed black magic), and in the README we have a mapping from those to common e-readers. Which ratios are the most common ones?

mikeshulman:
The pages I linked to before suggest that the most common ratios are 1.778 (16:9), 1.6 (16:10 = 8:5), and 1.333 (4:3 = 8:6), with some other possible ones being 1.3 (13:10), 1.3.24 (8.94:6.75), 1.356 (8:5.9), and 1.706 (8:4.69).  @liyishuai, what is the aspect ratio of your kindle?

liyishuai:
Mine is 1.324. I think three ratios should be convenient for everyone, say, 1.3, 1.5, and 1.7-ish.

mikeshulman:
Unless anyone can point to a particular device that has a ratio of 1.5-ish, I would suggest we stick to ratios like 1.333, 1.6, and 1.778 that match the most common actual devices.

liyishuai:
| Status | Format | Ratio | Height | Width |
|:---------:|:---------------|:-------|:------:|:-----:|
| Proposing | `ebook-wide` | 1.3125 | 6.3 | 4.8 |
| Existing | `ebook` | 1.6458 | 7.9 | 4.8 |
| Proposing | `ebook-narrow` | 1.7708 | 8.5 | 4.8 |

Makefile
```diff
@@ -36,6 +36,7 @@ OPTFILES = opt-letter.tex \
 	   opt-a4.tex \
 	   opt-ustrade.tex \
 	   opt-ebook.tex \
+	   opt-ebook-wide.tex \
```
mikeshulman: Should probably add `opt-ebook-narrow.tex` here too?

mikeshulman:
That's okay with me, although I'd probably lean towards 6.4x4.8 (1.333) for the wide.

JasonGross:
I am sold by the images in https://github.com/HoTT/book/pull/1037#issuecomment-530907614 which argues for 6.3 rather than 6.4 for the height

mikeshulman:
The image shows that 6.3 is better for @liyishuai's particular model of Kindle.  But I don't see any reason to believe that that model is representative; my brief Internet seach suggested that there are many different models of Kindle with different aspect ratios.  (Although if you know differently, please educate me!)  And the Internet also tells me that iPads are 4x3 (1.333).

liyishuai:
It'll be nice to have an iPad user to show us the difference. If 6.4 fits perfectly I'm happy to change.

JasonGross:
I would lean towards erring on the wider side; often my issue with pdfs on Kindle is that the text is too small.  If the page is a bit too short, then I have to turn the page slightly more often, which is not a big deal.  If the page is too tall, Kindle will shrink it and make the text smaller.  (My experience is also that many Kindles put some sort of text at the bottom, so there is less space in general than the actual aspect ratio of the screen.)

PR #1040, opened 2019-12-26
Update .travis.yml

spitters:
This should fix
https://travis-ci.org/HoTT/book/jobs/629638947

As per 2.11:
https://www.ctan.org/tex-archive/macros/latex/contrib/titlesec/

PR #1042, opened 2020-01-21
Don't use undefined term "boundedness"

### LINKED ISSUE #1041: Exercise 11.2 refers to a term "boundedness" which is not defined
Exercise 11.2 says "Suppose we remove the boundedness condition in Definition 11.2.1". However, Definition 11.2.1 does not use the word boundedness anywhere. My first idea about a fix is to just change "the boundedness condition" to "the <I>inhabited</I> condition". I open this as an issue because (1) maybe there is another better way to word this, and (2) is it important to refer to "boundedness" as a term or a concept?

With respect to (2) I notice that [BT09] says "A real number represented as a cut is therefore a pair of subsets, i.e. an element of P(Q) × P(Q), that is rounded, bounded (i.e. inhabited), disjoint and located." but I'm not sure whether we prefer (1) to choose one term and don't try to go too deeply into the words that other works use for the same concept, or (2) to mention other terms which are "sufficiently widely" used.

jkingdon:
~Although other works use "boundedness" as a synonym for "inhabited"
in this definition, we just call it "inhabited".~ The reference to boundness will work better if we define that term and also reference the condiiton more specifically by number.

Fixes #1041 . There's slightly more discussion in that issue, ~but this pull request provides the simplest solution, of just changing the exercise to match the terminology in the definition.~


reals.tex
```diff
@@ -3072,7 +3072,7 @@ \section{The surreal numbers}
 
 \begin{ex} \label{ex:RD-extended-reals}
   %
-  Suppose we remove the boundedness condition in \cref{defn:dedekind-reals}.
+  Suppose we remove the \emph{inhabited} condition in \cref{defn:dedekind-reals}.
```
DanGrayson: The phrasing is not the best, because it's not the "inhabited condition" that is removed, but "the condition that asserts that the upper and lower sets are inhabited" that is removed.

jkingdon:
Travis is failing with "mv: cannot stat 'titlesec/*.tss': No such file or directory". I don't know if there is anything I can or should do about that.

reals.tex
jkingdon: > The phrasing is not the best, because it's not the "inhabited condition" that is removed, but "the condition that asserts that the upper and lower sets are inhabited" that is removed.

By analogy to Exercise 11.4 I suppose another option is "the inhabitedness condition" or even "both inhabitedness conditions".

I guess I don't object to "the condition that asserts that the upper and lower sets are inhabited" although I guess we usually try to be more concise? Something like "the condition that the upper and lower cuts are inhabited" is slightly shorter.

I'm sold on at least making it "inhabitedness" (without italics) instead of "inhabited" but from the possibilities above I'm not sure I have an opinion.

DanGrayson: "inhabitedness" sounds bad to my ear

DanGrayson:
Once #1040 is pulled, travis should succeed.

jkingdon:
I've updated this with the suggested wording from https://github.com/HoTT/book/pull/1042#issuecomment-576524565 which seems to be the best solution so far.

jkingdon:
> Once #1040 is pulled, travis should succeed.

Ah, thanks for the pointer. I'll wait for things to get resolved over there before doing anything about travis.

mikeshulman:
I'm glad you like my suggestion.  One further suggestion is that one should never refer explicitly to the number of an item like "(i)", instead the item should be given a `\label` and the reference should be a `\ref`.

jkingdon:
> I'm glad you like my suggestion. One further suggestion is that one should never refer explicitly to the number of an item like "(i)", instead the item should be given a `\label` and the reference should be a `\ref`.

Ah, thanks for the pointer about `\label` and `\ref` - that was easier than I thought. I've pushed up that change and it looks better (for example, is clickable, at least in my pdf viewer).

mikeshulman:
Well, #1040 was pulled, but Travis is still failing on this PR (I restarted it to be sure after merging #1040).

DanGrayson:
Inexplicable.  I've restarted one of the jobs, just to see.

DanGrayson:
Okay, that didn't work, so I've merged the master branch into the pull request branch.

reals.tex
```diff
@@ -177,7 +177,8 @@ \section{Dedekind reals}
   : \Q \to \Omega$ which is:
   %
   \begin{enumerate}
-  \item \emph{inhabited:} $\exis{q : \Q} L(q)$ and $\exis{r : \Q} U(r)$,
+  \item \label{defn:dedekind-reals-inhabited}
+    \emph{inhabited (i.e. bounded):} $\exis{q : \Q} L(q)$ and $\exis{r : \Q} U(r)$,
```
DanGrayson: after each "i.e." should come a comma

jkingdon: > after each "i.e." should come a comma

I've pushed up a commit to add a comma after "i.e."

mikeshulman: Or a backslash.  Or replace the space by a tilde.

PR #1048, opened 2020-04-03
Improve typography, using true small capitals and microtypography

dbenjaminmiller:
This is a typographic adjustment to the book, and does not impact the content.

There are two changes which are:

(1) Use Domitian as the body text font (the math font remains Pazo). Domitian is an updated and extended version of the URW Palladio font (which was the version of Palatino being used already). Its basis is an updated version of Palladio released by URW in 2015; Domitian goes further by adding true small capitals. Now, you will notice that the main difference introduced by this change is the presence of the small capitals in the headers. This is in contrast with the formerly-used package, which used scaled-down versions of the normal-size capitals instead. This is a bit ugly and is not generally considered typographically correct where it can be avoided. (Disclaimer: I am the maintainer of this package.)

(2) I have enabled microtypography, which is a feature of pdfLaTeX (and has been for a number of years). Two features of microtypography have been enabled: protrusion and expansion. The first of these features, protrusion, means that punctuation may slightly protrude from the edge of the line; this is common in professionally published material. Expansion, meanwhile, allows for characters to be stretched and squeezed very slightly, in a way which is not visually noticeable. The benefit of these small adjustments is that the lines are spaced in a way which looks more natural, with far fewer line breaks in the middle of words and the avoidance of bad compositional features such as rivers. Also, I have set `microtype` to use protrusion settings specifically tailed for Palatino.

These packages are part of the standard TeX distributions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,13 +17,13 @@ and nightly builds are available on the
 ## Prerequisites and compilation
 
 To compile the book for yourself you need a fairly new version of LaTeX.
-[Texlive](http://www.tug.org/texlive/) 2012 is confirmed to work. You might need
+[Texlive](http://www.tug.org/texlive/) 2018 is confirmed to work. You might need
 to install some packages; see `main.tex` for packages that are used by the book.
 
 [BasicTeX](http://www.tug.org/mactex/morepackages.html), which is a minimalistic
 version of MacTeX, is confirmed to work once the following packages have been
-installed: `tlmgr`, `install`, `braket`, `comment`, `courier`, `enumitem`,
-`helvetic`, `mathpazo`, `nextpage`, `ntheorem`, `palatino`, `rsfs`, `stmaryrd`,
+installed: `tlmgr`, `install`, `braket`, `comment`, `courier`, `domitian`, `enumitem`,
+`helvetic`, `mathpazo`, `microtype`, `nextpage`, `ntheorem`, `palatino`, `rsfs`, `stmaryrd`,
 `symbol`, `titlesec`, `wallpaper`, `wasy`, `wasysym`, `xstring`, `zapfding`.
 
 You also need the `make` utility. The book is a fairly complex piece of LaTeX
--- a/main.tex
+++ b/main.tex
@@ -56,10 +56,14 @@
 \usepackage{multicol}
 
 %%% Set the fonts
-\usepackage{mathpazo}
+\usepackage{mathpazo,domitian}
 \usepackage[scaled=0.95]{helvet}
 \usepackage{courier}
 \linespread{1.05} % Palatino looks better with this
+
+%%% Use microtypography
+\usepackage[expansion=all,protrusion,step=1]{microtype}
+\DeclareMicrotypeAlias{Domitian}{ppl}
 
 \usepackage{ifpdf}
 


dbenjaminmiller:
Er, actually apparently there is one issue I need to fix. Hold on. There is an issue with the Domitian support file in the present version, evidently, and I need to fix that. After that I will reopen this PR.

mikeshulman:
At #584 the consensus seemed to be that it's not worth trying to make changes like this because it would require re-proofreading the entire book (half a dozen times, one for each edition).

dbenjaminmiller:
Thanks. I realize I had one glitch in my Domitian support file (which made the small capitals not work with f-ligatures properly). I'm fixing that, uploading it to CTAN and then it'll be updated in TeX Live 2020 (imminently). I suggest that then, I would reopen a new commit only changing that font setting but not enabling microtype until the new edition, if that's all right

mikeshulman:
Well, even for changing fonts I would be worried that proofreading would be required.  Any opinions from the other participants in #584 (@andrejbauer, @peterlefanulumsdaine )?

EgbertRijke:
I also recall that we decided not to use microtype, because it would require another round of proofreading. The same goes for changing the font. I do agree that microtype is a good package, so if we ever decide on a new edition (which will require proof reading anyway), then we should include it.

peterlefanulumsdaine:
I tried this out (both the font switch and enabling microtype, separately), with a quick visual diff to see how big the changes were.  In each case, the changes are non-trivial: many paragraphs re-flow, and sometimes gain or lose a line, and some page breaks shift slightly.  So these are definitely big enough to require careful re-proofing — let’s leave them till the 2nd ed.

@dbenjaminmiller Thank you for the contribution in any case — these are nice improvements, and it’ll be good to have them for the 2nd ed!

peterlefanulumsdaine:
Actually, cleaning up some other outstanding PR’s/issues, I realise I’ve had a bit of a double standard.  The main dangers I’ve been worrying about here and in #584 (e.g. widows/orphans/bad diagram placement due to page-breaks shifting) could just as well arise from the changes in other PR’s (e.g. adding a sentence to the text) — but I haven’t been checking carefully for such knock-on effects when merging other PR’s.

Have the other maintainers been checking this more scrupulously than me?  Or if not, should we reconsider these typographical improvements?  The two kinds of problem I can see they might cause are:

- overfull/underfull hboxes: these are easily found by checking the build log;
- warts due to page-breaks shifting as described above: if we haven’t been checking for these when merging other PR’s, I don‘t think it makes sense to lose sleep over them here.

@mikeshulman and other maintainers — am I missing any other kind of problems that could occur?

mikeshulman:
No, I admit I haven't been checking for that sort of thing either.  The only difference I can see between that and a change like this is that this change affects the entire book at once, so there's more potential for problems to occur.

Personally, I am not a typography geek and so I don't really see the need for a change like this.  But if other maintainers think it would be worthwhile and the potential problems aren't any worse than those arising from other changes, I won't object.

dbenjaminmiller:
Thanks for the input --- the one thing I'd note is that in my Domitian support package, at the moment, there is only glitch: in small caps, ligatures get erroneously inserted: like AWORDCONTAININGfiLIKETHIS ... I will fix that and push it to CTAN, but it won't take hold except in TL2020, which comes out in April. So if you're going to apply this change in the meantime, only do the microtype line.

dbenjaminmiller:
For now, though, it should be fine to just add this one line: `\usepackage[expansion=all,protrusion,step=1]{microtype}`.

But I will note that it is extremely rare that `microtype` introduces a typographical issue. But let me see what the result is, just by enabling `microtype` only, and see if the overfull/underfull boxes are any different…

dbenjaminmiller:
Based on the counts of overfull and underfull boxes in the log, for `make hott-ebook.pdf`, we go from 321 to 119 overfull, and 48 to 36 underfull, just by adding `\usepackage[expansion=all,protrusion,step=1]{microtype}` in `main.tex`.

PR #1050, opened 2020-04-04
Better nightly/ file deletion criterion

### LINKED ISSUE #986: keep old nightly builds around
It seems that the old nightly builds are not being kept around long enough.  They need to be kept around for at least a day, I think, since the redirects are only updated daily.  Was this a change due to the recent fix to make the nightly builds work again?

JasonGross:
Rather than using heuristics, we just directly ask if the file is the
one pointed to by saunders.

Closes #986, hopefully

### DIFF
--- a/etc/ci/update_nightlies.sh
+++ b/etc/ci/update_nightlies.sh
@@ -64,16 +64,19 @@ git checkout -b gh-pages upstream/gh-pages || exit 1
 
 mkdir -p nightly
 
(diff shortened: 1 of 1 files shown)


mikeshulman:
Looks worth a try, thanks.  Should we just merge it and see what happens?

mikeshulman:
Hmm, looks like we neglected to actually merge this.

PR #1052, opened 2020-04-13
Remark about context well-formedness metatheory; close #1051

### LINKED ISSUE #1051: Slight contradiction between sections A.2.1 and A.2.2
The end of section A.2.1 says

> It is a meta-theoretic property of the system that if
> x 1 :A 1 , . . . , x n :A n |-- b : B is derivable, then the context ( x 1 :A 1 , . . . , x n :A n )
> must be well-formed; thus ctx- EXT does not need to hypothesize well-
> formedness of the context to the left of x n .

However section A.2.2 starts with rule Vble which has such a well-formed context hypothesis.

mikeshulman:


DanGrayson:
lgtm

PR #1063, opened 2020-09-03
Small fix for the proof of Theorem 11.5.6

phi16:
- typo: $n$ → $k$
- include $i=k$ for the case $y=1$ which doesn't have $i$ for $0 \leq i < k$ satisfying $(i-1)/k < y < (i+1)/k$
- remove application of $r$ (the equality looks using the fact $r$ behaves just as identity now, it looks somehow typo?)

And sorry for not writing eratta, but I don't know how to write because it includes several points (could I write it for each point?).


mikeshulman:
This looks good to me; thanks.  I think it would be fine to have either one erratum for all of them or one for each change.  If you'd be willing to add errata that would be great!

phi16:
Thank you for comment, I added a single and simple erratum.

PR #1066, opened 2020-10-17
Fix anomalous newline screwing things up

pthariensflame:
Prior to this fix, `version.tex` would have the following contents:
```latex

ewcommand{\\OPTversion}{
```
This would cause errors when building.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -107,7 +107,7 @@ log-check:
 	! grep -n "!! Input index error" hott-online.ilg /dev/null
 
 version.tex:
-	/bin/echo -n '\newcommand{\OPTversion}{' > version.tex
+	/bin/echo -n '\\newcommand{\\OPTversion}{' > version.tex
 	git describe --always --long >> version.tex
 	/bin/echo -n '}' >> version.tex
 
--- a/Makefile
+++ b/Makefile
@@ -107,9 +107,7 @@ log-check:
 	! grep -n "!! Input index error" hott-online.ilg /dev/null
 
 version.tex:
-	/bin/echo -n '\\newcommand{\\OPTversion}{' > version.tex
-	git describe --always --long >> version.tex
-	/bin/echo -n '}' >> version.tex
+	printf '\\newcommand{\\OPTversion}{%s}\n' "`git describe --always --long`" > version.tex
 
 # these warnings are mostly spurious, and could have been prevented by a better makeindex algorithm
 log-check-for-warnings:


DanGrayson:
The command currently there seems correct:
```sh
$ /bin/echo '\newcommand{\OPTversion}{' > version.tex
$ cat version.tex 
\newcommand{\OPTversion}{
```

pthariensflame:
It's also failing on CI, but the change is what *fixes* the build on my own machine.  I wonder if it's OS-dependent, since I'm on a Mac.

mikeshulman:
It's shell-dependent.  The version currently in the makefile is what works in bash, while your version is what works in tsch.

DanGrayson:
This experiment shows that bash and tcsh respond the same way, but sh and zsh do not:
```sh
einsteinium$ bash
einsteinium$ echo '\\'
\\
einsteinium$ sh
sh-3.2$ echo '\\'
\
sh-3.2$ csh
[einsteinium:~/src/speedily-dictionary.git] dan% echo '\\'
\\
[einsteinium:~/src/speedily-dictionary.git] dan% tcsh
[einsteinium:~/src/speedily-dictionary.git] dan% echo '\\'
\\
[einsteinium:~/src/speedily-dictionary.git] dan% zsh
einsteinium% echo '\\'
\
```

On the other hand, if I try various shells with this makefile:
```make
a :
	echo '\\' > /tmp/asdf
	cat /tmp/asdf
```
, then I always get just one \ in the output.

If it really is shell dependent, then we could add a line like this one to the makefile:
```make
SHELL := /bin/sh
```

mikeshulman:
The issue is the treatment of `\n`, not `\\`:

```
shulman@sahadeva:~$ bash
shulman@sahadeva:~$ echo '\n'
\n
shulman@sahadeva:~$ exit
shulman@sahadeva:~$ tcsh
sahadeva:~> echo '\n'


sahadeva:~> exit
```

flexibeast:
@DanGrayson:

> SHELL := /bin/sh

This won't necessarily do what you intend, as on some systems, `/bin/sh` is just a symlink to the actual binary. So depending on the OS and user configuration, it could point to any of e.g. `dash`, `bash`, `tcsh`, `zsh`, `ksh`, etc.

mikeshulman:
It appears to be even worse than that.  As far as I can tell, at least on my machine, setting `SHELL=/bin/sh` in a makefile doesn't actually invoke `/bin/sh`.  On my machine the latter is a symlink to `/bin/dash`, which behaves like tcsh and zsh when testing manually; but setting `SHELL=/bin/sh` in a makefile doesn't change the behavior -- the makefile still behaves like bash.  (Setting `SHELL` to `/bin/tcsh` or `/bin/zsh` does change the behavior.)

Maybe the best solution is to set `SHELL=/bin/bash` in the makefile (and leave everything else unchanged)?  Does that fix your problem @pthariensflame ?

pthariensflame:
@pthariensflame It does not fix it.  Notably, I'm *already* using bash 5.0.18 as my default shell.

pthariensflame:
Also cross-checked GNU make *vs.* BSD make; that doesn't change the behavior for me.

pthariensflame:
The following extremely dumb solution worked, and *should* work in any vaguely POSIX-ish shell, but I'm wary of asserting that strongly until others test it!

```diff
diff --git a/Makefile b/Makefile
index bf487a4..a020f3b 100644
--- a/Makefile
+++ b/Makefile
@@ -107,8 +107,9 @@ log-check:
 	! grep -n "!! Input index error" hott-online.ilg /dev/null
 
 version.tex:
-	/bin/echo -n '\newcommand{\OPTversion}{' > version.tex
-	git describe --always --long >> version.tex
+	/bin/echo -n "\\" > version.tex
+	/bin/echo -n 'newcommand{\OPTversion}{' >> version.tex
+	git describe --always --long | tr -d "\n" >> version.tex
 	/bin/echo -n '}' >> version.tex
 
 # these warnings are mostly spurious, and could have been prevented by a better makeindex algorithm
```

SkySkimmer:
The shell shouldn't matter since `/bin/echo` is used instead of builtin `echo`.
builtin `printf` may be more convenient for precise prints.

Makefile
```diff
@@ -107,7 +107,7 @@ log-check:
 	! grep -n "!! Input index error" hott-online.ilg /dev/null
 
 version.tex:
-	/bin/echo -n '\newcommand{\OPTversion}{' > version.tex
+	/bin/echo -n '\\newcommand{\\OPTversion}{' > version.tex
```
JasonGross, line 110: I've run into this issue in my own projects recently on Windows, and the solution I've used is
```suggestion
	printf '\\newcommand{\\OPTversion}{' > version.tex
```

DanGrayson:
@pthariensflame -- please tell us what OS and version of it you have, so we can reproduce the problem.

mikeshulman:
@SkySkimmer On my machine, at least, the shell does matter.  I presume the command line in the makefile is parsed by the shell before being passed to the program `/bin/echo`.

@pthariensflame Note that `make` does [not](https://www.gnu.org/software/make/manual/html_node/Choosing-the-Shell.html) take the variable `SHELL` from the environment, so your default shell for interactive use doesn't affect it.  Did you try setting `SHELL` in the makefile itself?

mikeshulman:
@pthariensflame regarding your "dumb solution", I am mystified by the need to break the line into two.  In what conceivable world would `echo "\\n"` lead to a newline?

Apparently, the world of `zsh`:
```
$ zsh
~ % echo "\\n"


~ % echo "\\"
\
~ % echo "\\\\"
\
~ % echo "\\\\\\\\"
\\
```
`dash` behaves the same way.  What on earth are they doing?  The first three of these suggest some kind of "macro expansion", where the first backslash quotes the second backslash, and then the resulting single backslash potentially combines with the following character.  But aside from the fact that this is kind of the opposite of what backslash *quoting* a character is supposed to mean (*preventing* it from having a special effect), a true macro expansion (of unlimited depth, at least) would make the octuple-backslash also yield a single one.

Whatever it is, this can't be POSIX-compliant... can it?

mikeshulman:
even more bizarre:
```
~ % echo "\\\n"
\n
~ % echo "\\\""
\"
```
The second one is what I would expect in principle, but seems inconsistent with the behavior of `\\\\`.  The first one I can't square with anything.

DanGrayson:
> @pthariensflame regarding your "dumb solution", I am mystified by the need to break the line into two. In what conceivable world would `echo "\\n"` lead to a newline?
> 
> Apparently, the world of `zsh`:
> 
> ```
> $ zsh
> ~ % echo "\\n"
> 
> 
> ~ % echo "\\"
> \
> ~ % echo "\\\\"
> \
> ~ % echo "\\\\\\\\"
> \\
> ```
> 
> `dash` behaves the same way. What on earth are they doing? The first three of these suggest some kind of "macro expansion", where the first backslash quotes the second backslash, and then the resulting single backslash potentially combines with the following character. But aside from the fact that this is kind of the opposite of what backslash _quoting_ a character is supposed to mean (_preventing_ it from having a special effect), a true macro expansion (of unlimited depth, at least) would make the octuple-backslash also yield a single one.
> 
> Whatever it is, this can't be POSIX-compliant... can it?

Processing here happens in two stages: first the command line is parsed, and each `\\` is converted to `\`.   Then the shell's built-in "echo" command does its processing, which *apparently* involves the same conversion, but leaves spurious escape sequences, such as a lone `\` or a `\ ` unchanged.

DanGrayson:
> even more bizarre:
> 
> ```
> ~ % echo "\\\n"
> \n
> ~ % echo "\\\""
> \"
> ```
> 
> The second one is what I would expect in principle, but seems inconsistent with the behavior of `\\\\`. The first one I can't square with anything.

The model I have in mind explains this: command line processing converts each \\ to \, and then the echo command does not further processing.

mikeshulman:
Ah, I see.  But that means that since the makefile is invoking `/bin/echo` instead of the shell builtin `echo`, it should work with all shells to just say `/bin/echo -n "\\newcommand..."`.

DanGrayson:
> Ah, I see. But that means that since the makefile is invoking `/bin/echo` instead of the shell builtin `echo`, it should work with all shells to just say `/bin/echo -n "\\newcommand..."`.

Maybe ...

Clearer would be  `/bin/echo -n '\newcommand...'`, to avoid the command line conversion of \\ to \.

mikeshulman:
> Clearer would be `/bin/echo -n '\newcommand...'`, to avoid the command line conversion of \\ to \.

That's what the current Makefile has.  And it works on all the shells I can test, but @pthariensflame says it doesn't work for him.  So maybe I was wrong and it's not about the shell at all.

pthariensflame:
@DanGrayson I'm running macOS Big Sur beta 10; I gave my shell version earlier.  I'll check what versions of BSD and GNU make I have in a bit.

Makefile
```diff
@@ -107,7 +107,7 @@ log-check:
...
 	git describe --always --long >> version.tex
 	/bin/echo -n '}' >> version.tex
```
JasonGross, line 112: Another option:
```suggestion
	printf '\\newcommand{\\OPTversion}{%s}\n' "`git describe --always --long`" > version.tex
```

JasonGross:
According to https://stackoverflow.com/a/47253034/377022, 
> using `\n` with `echo` is unpredictable; some versions of `echo` convert it to newlines, some don't. It's actually changed between different versions of OS X.

According to https://unix.stackexchange.com/a/219274/45323,

> An `echo` implementation which strictly conforms to the Single Unix Specification will add newlines if you do:
> 
> ```shell
> echo 'line1\nline2'
> ```
>
> But that is not a reliable behavior. In fact, there really *isn't* any standard behavior which you can expect of `echo`.

DanGrayson:
So switching to `printf` seems called for.

mikeshulman (approved):
So it's not actually about the shell, but about the `echo`!  You learn something new every day.

PR #1069, opened 2020-12-09
Clarify that the universe U is itself a type

GeoffChurch:
I thought the universe U might be a different sort of object (like a collection instead of a set) until I read "Like any type" in the next sentence.

### DIFF
--- a/introduction.tex
+++ b/introduction.tex
@@ -124,7 +124,7 @@ The univalence axiom also further strengthens the homotopical view of type theor
 \subsection*{Univalent foundations}
 
 Very briefly, the basic idea of the univalence axiom can be explained as follows.
-In type theory, one can have a universe $\UU$, the terms of which are themselves types, $A : \UU$, etc.
+In type theory, one can have a \emph{universe} type, $\UU$, the terms of which are themselves types, $A : \UU$, etc.
 Those types that are terms of $\UU$ are commonly called \emph{small} types.
 \index{type!small}%
 \index{small!type}%


martinescardo:
The new sentence says "universe type" but I think it should be "type universe" to follow common parlance.

martinescardo:
But perhaps it would be better to completely rewrite the sentence.

mikeshulman:
I think "universe type" sounds just fine.  Don't we talk about "function types", "pair types", "natural numbers types", etc.?  I suppose to be really consistent with those it should be "type type", but ugh.

mikeshulman:
The point of saying "universe type" is that the universe itself *is* a type.  In "type universe" the word "type" refers rather to the *elements* of the universe.

mikeshulman:
However, I'd be happy to hear suggestions for better ways to rewrite the sentence.

DanGrayson:
Maybe something like "In type theory, one can have a type whose elements are themselves types; such a type is called a \emph{universe} and is usually denoted by $\UU$."

RobertHarper:
This sounds the most appealing to me, and is often the way I describe them.  Such types whose elements are types need not be “universal”, though; the empty type is a universe, for example.  Nevertheless, I think that the formulation Dan proposes is good.

(c) Robert Harper.  All Rights Reserved.

mikeshulman:
Thanks Dan; see #1076.

DanGrayson:
I think the empty type is not a universe.  If it were, then in a context where a variable X has been declared to be an element of the empty type, we could use it as a type, adding, for example, a variable f of type X->X to the context and verifying that the new context is well-formed.

RobertHarper:
Good point, I meant only that one could have an empty universe.

(c) Robert Harper.  All Rights Reserved.

PR #1072, opened 2021-02-04
Reword first sentence of proof of Lemma 10.3.22, close #1071

### LINKED ISSUE #1071: Set-truncation in Lemma 10.3.22
In the statement of Lemma 10.3.22, should X be assumed to be a set?  If not, then in the proof, B is defined as a quotient of the (non-set) Sigma_{x : X} Fx, so there should probably be an explicit set-truncation.  If X is assumed to be a set, should the proof say "set quotient" instead of "quotient"?

jdchristensen:
Refer to Remark 6.10.1, since the proof takes the set-quotient of a type that may not be a set.


mikeshulman:
Thanks!  I wonder whether this needs an erratum?

jdchristensen:
I don't think so.  It's just more clearly expressing what was probably intended all along.

PR #1077, opened 2021-02-27
Present set theory as type theory with one type

VincentSe:
This pull request explains that the 2 layers of set theory come from the unique type available there: set.

### DIFF
--- a/preliminaries.tex
+++ b/preliminaries.tex
@@ -11,11 +11,11 @@ However, it behaves differently from set theory in several important ways, and t
 Explaining these differences carefully requires us to be more formal here than we will be in the rest of the book.
 As stated in the introduction, our goal is to write type theory \emph{informally}; but for a mathematician accustomed to set theory, more precision at the beginning can help avoid some common misconceptions and mistakes.
 
-We note that a set-theoretic foundation has two ``layers'': the deductive system of first-order logic,\index{first-order!logic} and, formulated inside this system, the axioms of a particular theory, such as ZFC.
+Informally, there is only one type in set theory: set. According to set theory the natural number 2 is a set, the real exponential function is a set, and a line in euclidean space is also a set. All mathematical objects are interpreted as sets, except propositions and proofs.
+To accomodate propositions, a set-theoretic foundation has two ``layers'': the deductive system of first-order logic,\index{first-order!logic} and, formulated inside this system, the axioms of a particular theory, such as ZFC.
 Thus, set theory is not only about sets, but rather about the interplay between sets (the objects of the second layer) and propositions (the objects of the first layer).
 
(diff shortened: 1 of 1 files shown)


DanGrayson:
I think I can see why this rephrasing has been proposed, but the new version has its own problems.

"All mathematical objects are interpreted as sets, except propositions and proofs."  

In set theory, we don't think of propositions and proofs as mathematical objects.

"By contrast, having different types allows to reintegrate propositions and proofs into the same layer as the other mathematical objects."

It should be stated that we are contrasting type theory with set theory.

The phrase "different types" conveys no more information than the phrase "types", because "types" is plural.

The use of "reintegrate" rather than "integrate" implies that propositions and proofs were formerly (before ZFC) part of the same layer as the mathematical objects.

The verb "allows" is transitive and takes an object.

Saying "as the other mathematical objects" here implies that propositions and proofs are mathematical objects.

"The number 2 becomes of type ``natural number'', the Pythagorean theorem of type ``proposition'' and its proof of type ``Pythagorean theorem''. Therefore type theory is its own deductive system: it need not be formulated inside any superstructure, such as first-order logic."

Using "therefore" here is a bit too strong -- it sounds like something is being proven, but rather, the conclusion is just being illustrated.  The original version put that concluding statement first, so it sounds more forceful.

mikeshulman:
> I think I can see why this rephrasing has been proposed

I can't.  What are you trying to achieve here?

awodey:
I don’t think the original is in need of improvement, and anyway this is not an improvement.

VincentSe:
The aim of this pull request is to ease the transition from set theory to type theory for classical mathematicians. I agree with all @DanGrayson remarks and can integrate them in the PR.

mikeshulman:
But exactly what difficulty have you encountered in that transition that you are trying to ease?

VincentSe:
I hear from classical mathematicians that type theory is so different from set theory that they don't even try to read it. Albeit simplified, presenting set theory as a type theory with just one type seems a powerful motive to switch to type theory. Even more so by showing that the logical layer in set theory was just a backdoor to bring back propositional types. In other words, classical mathematicians wanted types, but they didn't know it.

mikeshulman:
Why don't you write a little note or blog post of your own explaining that point of view?

VincentSe:
Well, probably for the same reason that you and others wrote the HoTT book, instead of writing scattered blog posts about HoTT.

telephon:
As far as I can tell, in ZFC-style set theory, there is no such thing as a type. Sets are not even explicitly defined. I would find it rather misleading to say that no type is the same as one single type.

awodey:
A good place for this sort of discussion is homotopytypetheory@googlegroups.com, or in the comments following a blog post on homotopytypetheory.org, or on the new HoTT Zulip.  But not in the HoTT book, which - except for typos and corrections of proofs - is pretty settled.

PR #1078, opened 2021-04-01
Udate Lulu.com book dimensions.

andrejbauer:


mikeshulman:
Why did the dimensions need to change?

Issue #1074, opened 2021-02-09
Example 9.1.15: every isomorphism is an identity?


giorgio93p:
I have trouble understanding what is meant by 

> If A is a category, then A_0 is a set if and only if for any a, b : A_0 , the type a ≅ b
> is a mere proposition. This is equivalent to saying that every isomorphism in A is an identity;
https://github.com/HoTT/book/blob/a0040d8c075d033f748508b09728fe6869ecc60d/categories.tex#L204

What does it mean for an isomorphism `f : a ≅ b` to _be_ an identity? Of course, it can't be equal to an identity in general, since an identity would have a different type altogether, e.g. `a ≅ a` (and `a`, `b` may not be judgementally equal). However, if we interpret "being an identity" as "being able to be transported to an identity", then this is true for isomorphisms in all categories, by equation (9.1.10) (setting `p` to be `refl_a` and `q` to be `isotoid(f^-1)`).

As an example, consider the discrete category on the interval type. In what sense is `seg` an identity?

mikeshulman:
Maybe it would be better to say "every *automorphism* is an identity".

DanGrayson:
Maybe expand that to "every automorphism is an identity arrow".

mikeshulman:
Sure.

PR #1097, opened 2022-01-24
Fix issue #1074

giorgio93p:
This PR fixes issue https://github.com/HoTT/book/issues/1074

### DIFF
--- a/errata.tex
+++ b/errata.tex
@@ -761,7 +761,7 @@ While the page numbering may differ between copies with different version marker
   % Chapter 9
   %
   \cref{ct:gaunt}
-  & merge of 1093602
+  & % merge of 1093602
   & Stating that every isomorphism is an identity is not very accurate (consider the discrete category on the interval type): a more accurate statement is that every automorphism is an identity arrow. Notice that for precategories, this property must be combined with skeletality for the equivalence to hold.\\
   %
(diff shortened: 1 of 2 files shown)


mikeshulman:
Thanks!  But now that I look at this change in context, I think the second clause of this sentence also needs changing.  I'm not sure exactly what it should say, but "skeletality" doesn't really make sense for (univalent) categories in HoTT, while the statement "every automorphism is an identity" when interpreted in classical mathematics doesn't imply skeletality.  What about something like

> Given that A is a category, this is equivalent to saying that every automorphism in A is an identity arrow.  On the other hand, if A is a precategory such that A_0 is a set, then A is a category precisely if it is skeletal (any two isomorphic objects are equal) *and* every automorphism is an identity arrow.  Categories of this sort are sometimes called **gaunt** [BSP11].  The classical notion of "skeletal category" does not really generalize well beyond the case when A_0 is a set, unless one considers Definition 9.1.6 itself to be such.

errata.tex
```diff
@@ -760,6 +760,10 @@
   %
   % Chapter 9
   %
+  \cref{ct:gaunt}
+  & merge of 1093602
```
mikeshulman, line 764:
```suggestion
  & % merge of 1093602
```

PR #1098, opened 2022-02-23
github actions to replace travis script

Alizter:
Hopefully this fixes the nightly builds

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,7 +45,10 @@ jobs:
           texlive-science \
           texlive-metapost \
           texlive-xetex
-
+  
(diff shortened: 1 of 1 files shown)


mikeshulman:
Hooray!  Thanks.  Is there any way to test this aside from merging it and seeing what happens?

Alizter:
I would have to modify the scripts since they are checking that HoTT/book is the branch. However, we can probably run the PR here. When you click on the "Actions" tab next to Pull Requests do you see a way to enable it?

mikeshulman:
I don't see any button to run the action, or any way to enable it that isn't already enabled.  But merging it is probably ok.

PR #1099, opened 2022-02-25
Use OAUTH_TOKEN from github repo

Alizter:
I have made a secret in this repo called OAUTH_TOKEN that the Github actions workflow will use. It currently contains my personal access token so when "HoTT Bot" tries to push in the scripts, it will be pushing using my authentication.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,6 +27,7 @@ jobs:
       UPDATE_ERRATA: ${{ matrix.update_errata }}
       UPDATE_NIGHTLIES: ${{ matrix.update_nightlies }}
       TARGETS: ${{ matrix.targets }}
+      OAUTH_TOKEN: ${{ secrets.OAUTH_TOKEN }}
 
     steps:
       # Install texlive and other deps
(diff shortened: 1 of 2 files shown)


Alizter:
Don't merge just yet something went wrong.

Alizter:
Alright rerunning worked. No idea what went wrong so probably just a networking issue.

favonia:
Theoretically, the access token built in GitHub Actions should be sufficient.

favonia:
Hi @Alizter, after extensive testing I came to the conclusion that we probably don't need your personal token. Also, I am rewriting these scripts from scratch, and wonder if you have any suggestions. The first PR of the effort is #1104.

Alizter:
I'm glad that you got it working. :-) I don't have any other suggestions. How did you avoid the use of a personal token? Also we can close this PR if you like.

favonia:
> @favonia I'm glad that you got it working. :-) I don't have any other suggestions. How did you avoid the use of a personal token? Also we can close this PR if you like.

1. GitHub Actions is equipped with a useful token by default. No need to use yours unless necessary (e.g., if you want to commit changes to _another_ repository). [Here is the documentation.](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#about-the-github_token-secret)
2. There might be other things in the existing scripts that might prevent your attempts from working.

Issue #1096, opened 2021-11-23
errata marking script broken?


mikeshulman:
Commit 2b9a7ef86 added an erratum with the correct syntax, but it doesn't seem to have been marked by the script.  Is the errata-marking script still running automatically?

(Separately, commit 518e3a95a9 added an erratum with incorrect syntax -- the comment character is missing.)

JasonGross:
I believe Travis is no longer a good option for CI for open source projects (see, e.g., [this blog post](https://earthly.dev/blog/migrating-from-travis/)), and if you want an open source project to run on them you need to pester them every so often to get them to give you more CI minutes.  I'd suggest that we port the CI scripts to GitHub actions.

JasonGross:
It looks like CI has not run since 1ac94087575f3aee19dcecc88ecdea2ba9170ca6 in February

mikeshulman:
Porting to github actions would be great.  Is anyone here able to do it?

JasonGross:
I can probably do it, if I find time.  The most non-obvious bit is figuring out what secret is encoded at https://github.com/HoTT/book/blob/1fc7194b9c9654c6d2c1299933403443467ff494/.travis.yml#L27-L29
(I think it's used for pushing the updated ebook to the wiki/gh-pages?) and either adding the underlying secret to https://github.com/HoTT/book/settings/secrets/actions or else generating a new secret and setting that up.

mikeshulman:
Any progress on this?  It's possible that secret has to do with updating the saunders server at cmu, which is supposed to include links with constant names to the files with varying names on the nightly build pages.

JasonGross:
Sorry, it fell off my todo list.  I'll try to get to this in the next couple of days.

mikeshulman:
Any progress?  It would be nice to have the rest of the CI working as well, e.g. the download links at homotopytypetheory.org/book are also broken and the nightly build page is out of date.

PR #1102, opened 2022-03-13
Rewrite CI scripts

favonia:
I was intrigued by #1096, #1098, and #1099 and decided to take a curious look. The current build system is a bit more complicated than I expected. Currently, I am cleaning up the scripts while learning them. This is still a WIP and, while I'm hopeful, I can't guarantee I would achieve something better in the end.

- [x] Rewrite updater of errata.tex
- [x] Rewrite the script for `make dvi`
- [x] Skip the testing of `make default` in the presence of `update-nightlies`
- [x] Rewrite and split `update-nightlies` into (1) wiki updater and (2) PDF updater
  - [x] Rewrite the wiki updater
  - [x] Rewrite the PDF updater; blocked by #1103
- [ ] Try other ways to install TexLive (optional)


etc/ci/add_upstream.sh
```diff
@@ -1,11 +1,10 @@
...
 # in case we're run from out of git repo
-DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
-pushd "$DIR" 1>/dev/null
+DIR="$(dirname "$(realpath "${BASH_SOURCE[0]}" )")"
+pushd "$DIR" 1>/dev/null || exit 1
+trap popd EXIT
 
 echo "Adding remote..."
-git remote add upstream git://github.com/HoTT/book.git
+git remote add upstream https://github.com/HoTT/book
```
JasonGross, line 9: Are you sure this doesn't break the authentication method?  What's the reason for this change?

favonia: The reason is that https://github.com/actions/checkout, by default, would not set up a SSH key.

favonia:
Updates: the `<meta>` redirection solution suggested in #1103 has been implemented in this PR. This is a major rewritea. One can check the PDF by following URLs like this:

> https://hott.github.io/book/hott-online.pdf.html

The above link would be redirected to https://hott.github.io/book/hott-online-1319-g54f404d.pdf.

favonia:
Some areas for improvements that I intend NOT to address in this PR:

1. Do people still care about DVIs? Can we drop the support for DVIs altogether?
2. Building these PDFs takes lots of time. Some thoughts:
   * Can the build system be made a bit more "parallel"? I tried forcing the make to run tasks in parallel and it did not end well.
   * Can we somehow restructure the LaTeX source so that it does not have to be rerun so many times?
5. Perhaps it would be helpful to find or even build a smaller texlive Docker image to reduce the build time.

mikeshulman:
Nice!  Is this ready to merge?

I don't suppose there's any way for the redirecting URLs to end in .pdf without the extra .html.

Did we ever support DVIs?  I don't remember ever offering them for download.

Does it really matter how long the build takes?  It only needs to be run when new changes are merged into the repository, which is quite infrequent.

favonia:
> Nice! Is this ready to merge?

Yes. We can just merge it now!

> Did we ever support DVIs? I don't remember ever offering them for download.

We check whether `make dvi` will run successfully every time, even if the generated DVIs are not used. I feel we could potentially remove all the code about DVIs, including the relevant code in `Makefile`. However, I intend not to make these changes in this PR even if we want to cut all the DVI-related code.

> Does it really matter how long the build takes? It only needs to be run when new changes are merged into the repository, which is quite infrequent.

It's mostly my perfectionism kicking in. Currently, the CI will also be run for every new commit in a pull request (but then the script will not push any changes to the website or the wiki unless it's run on the `master` branch). It is potentially useful to know whether some changes would break as soon as possible. I have tried to reduce the time to about 6-7mins by not building all the PDFs (that is, not running `make all` but only making the ones that will be offered for download). I'm more used to squeezing the CI time down to 3mins or less but 6-7mins in any case will not be the end of the world. So, in sum, it's probably okay.

mikeshulman (approved):

mikeshulman:
Thank you very much for doing this!!

favonia:
> I don't suppose there's any way for the redirecting URLs to end in .pdf without the extra .html.

Sorry I somehow missed this. It's actually already working. You can drop the `.html`. I was a bit hesitant because I am not sure whether GitHub will change this in the future and/or whether some browsers will rely on the file extension `.html` to correctly interpret the resource (they should not... but I don't have control over them). Shorter URLs are great to tell other people where to download the PDFs, but I would use longer versions on homotopytypetheory.org, for example.

- https://hott.github.io/book/hott-a4.pdf
- https://hott.github.io/book/hott-a4.pdf.html

mikeshulman:
Can we do something to prevent these actions from trying to run on forks?

favonia:
We can attach an `if` to a job instead of individual steps. However, I'd like to learn more about the purpose of stopping these actions. The updates in a forked repository will not affect the main repository (HoTT/book), so the correctness is maintained, and I thought it's good to know whether the CI _would have_ worked. So, I wonder if the concern is about the CI doing useless work? In sum, I wish to know (1) which actions to stop (updating gh-pages, updating wiki, or both) and (2) the motivation.

mikeshulman:
I think it makes sense to compile the book on pushes to forks, but not to try to update any posted versions.  I just pulled from upstream and pushed to my fork, and got a notification that pushing to the wiki failed (although not, interestingly, that pushing to gh-pages failed; maybe it actually pushed something to my personal gh-pages?).  I don't want everyone with a fork to get notifications like that every time they push.

favonia:
> I don't want everyone with a fork to get notifications like that every time they push.

Got it. I took a look and figured out what was going on: the code I used assumed that wiki pages already existed, but there are no wiki pages on your personal fork. The code was written by other people, and perhaps the easiest fix is to turn it off.

For gh-pages, yes, those files are in https://github.com/mikeshulman/book/tree/gh-pages

mikeshulman:
Thanks.  I don't think everyone with a fork should get versions of the book stored on their gh-pages either.

Issue #1103, opened 2022-03-13
Redirects for nightly builds


favonia:
I believe the redirect system set up in #799, #801, and #845 is no longer working. I played a bit with the CI scripts (see #1102) but got stuck because `saunders.phil.cmu.edu` is not working. I wonder what would be the plan to restore the links?

By the way, in addition to all the methods mentioned in #799, we could also put HTML files on GitHub with
```html
<meta http-equiv="refresh" content="0; url=real-file-123456.pdf">
```
to solve the problem. The advantage is that it does not involve other servers and the redirects will always be up to date. The downside is that people might not expect an HTML redirect page before downloading the actual PDF, and tools like `curl` and `wget` do not support this kind of redirects.

Yet another solution is to have a landing page directly on GitHub, say, at `https://hott.github.io/book/download.html`.

mikeshulman:
It would be nice not to need Saunders.

favonia:
It is easy to generate those HTML files with `<meta />` redirection. I wonder how we can make a decision on this.

favonia:
Closed by #1102.

favonia:
Related: #1105.

PR #1104, opened 2022-03-13
GitHub Actions to update errata.tex

favonia:
While we are deciding the fate of nightly builds in #1103, I believe we can already enable the updater for `errata.tex`. This script will be triggered by (1) manual execution, (2) any commit to the master branch, and (3) any pull request from the same repo (for example this one). The commit message will be `Mark Errata (auto)` as the one used by the original Travis script.

The code was extracted from #1102. I tried to minimize changes to other files in this PR.

~~PS: Currently, I [manually disabled this action](https://github.com/HoTT/book/actions/workflows/errata.yml) because otherwise it would _immediately_ try to update `errata.tex` in this PR.~~


mikeshulman:
Seems ok as far as I can tell.  Do we just have to merge it and see whether it works?

JasonGross:
> PS: Currently, I [manually disabled this action](https://github.com/HoTT/book/actions/workflows/errata.yml) because otherwise it would _immediately_ try to update `errata.tex` in this PR.

What's wrong with it updating the errata in this PR?

.github/workflows/ci.yml
```diff
@@ -56,9 +46,3 @@ jobs:
           
     - name: Build
       run: make ${{ matrix.targets }}
-    
-    - name: Update errata
-      run: etc/ci/update_errata.sh
- 
-    - name: Update nightlies
```
JasonGross, line 63: For this PR, we should leave in the updating of nightlies, right?

favonia:
> > PS: Currently, I [manually disabled this action](https://github.com/HoTT/book/actions/workflows/errata.yml) because otherwise it would _immediately_ try to update `errata.tex` in this PR.
> 
> What's wrong with it updating the errata in this PR?

I prefer updating it in the master branch instead of this PR because I didn't want to make the impression that this PR has directly changed errata.tex in any way. That said, this is a very minor point that I don't feel strongly about. There's nothing really "wrong".

.github/workflows/ci.yml
favonia: No, I think we should pause the updating until we have decided what to do and then completed the transition. The current code is already not working (well), and the redirect server is down. The updater had stopped working for more than one year and I believe we can wait a bit longer. Running the semi-working updater for nightly builds _now_ is not very useful IMO.

favonia:
> Seems ok as far as I can tell. Do we just have to merge it and see whether it works?

Yes. In fact, it already worked on this PR, but then I manually reverted the automatic commit.

favonia:
@JasonGross @mikeshulman Alright, I re-enabled the workflow so that, as you can see, the script managed to update `errata.tex` in this PR as 574a6d8251c4a6a29f687707d8958b85080eaee7. It is easy to change the detail of the automatic commit---including setting its author to the one who originally triggers the event. I still feel it is better to put a hold on updating nightly builds until #1103 is resolved.

mikeshulman (approved):

PR #1106, opened 2022-03-18
fix incorrectly prepared erratum

mikeshulman:
A missing `%` prevented the errata-marking script from catching this one.

### DIFF
--- a/errata.tex
+++ b/errata.tex
@@ -887,7 +887,7 @@ While the page numbering may differ between copies with different version marker
   & \eqref{eq:untruncated-linearity} should be $c:\prd{q, r : \Q} (q < r) \to (q < x) + (x < r)$, and therefore the use of $c$ in the proof should be $c(s,t)$ rather than $c(x,s,t)$.\\
   %
   \cref{analysis-interval-ctb}
-  & merge of ad72d4e
+  & % merge of ad72d4e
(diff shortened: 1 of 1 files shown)


JasonGross (approved):

favonia:
Should we re-enable the CI on this PR so that the script would automatically calculate the commit hash and update the errata?

mikeshulman:
won't it happen automatically when the PR is merged?

favonia:
> won't it happen automatically when the PR is merged?

Yes, it will. I got the impression from #1104 that we wished to run the scripts eagerly.

mikeshulman:
I don't see any reason to run them on PRs before they're merged.  The PR that was implementing the new CI was a special case, since in that case it was nice to actually test the CI before merging it.  But now that we trust the CI is working, its actions only need to be done on actual merge.

favonia:
Should we merge this now? I am curious about whether #1109 fixes everything.

mikeshulman:
Looks like the run failed on "Check if errata.tex is clean".

favonia:
> Looks like the run failed on "Check if errata.tex is clean".

Right. It failed because the second stage failed to pull in the commit that fixed the errata.tex. I will fix it later.

In the meanwhile, is it still important to check whether DVIs can be built? The code could be further simplified if we remove that part from the CI. Currently, the CI just tries to build those DVI files and then discard them.

mikeshulman:
I think we can ignore DVIs.

PR #1109, opened 2022-03-23
Fix errata marking

favonia:
Upon closer inspection of errata.tex, I noticed that the errata marking in #1104 has led to incorrect results, possibly due to how `./mark-errata` is designed and how GitHub's actions/checkout performs merge for pull_request events. (I did not really change the script in #1104.) My current hypothesis is that the script can only be run on the master branch, and the CI has been updated accordingly. @JasonGross

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,8 +12,8 @@ jobs:
     name: Update errata
     runs-on: ubuntu-latest
     # It seems the ./mark-errata script would not work correctly for pull_request events.
-    # We want to run the script on the master branch anyways.
-    if: ${{ (github.event_name == 'workflow_dispatch' || github.event_name == 'push') && github.ref == 'refs/heads/master' }}
+    # We want to run the script only on the master branch anyways.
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -11,8 +11,9 @@ jobs:
   update-errata:
     name: Update errata
     runs-on: ubuntu-latest
-    # it seems impossible to update a branch in a forked repo with the default token
-    if: ${{ ! (github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name != github.repository) }}
+    # It seems the ./mark-errata script would not work correctly for pull_request events.
+    # We want to run the script on the master branch anyways.
+    if: ${{ (github.event_name == 'workflow_dispatch' || github.event_name == 'push') && github.ref == 'refs/heads/master' }}
```
JasonGross, line 16: Do you also want to check that we're on the main repo here, and not run on people's forks' master branches?

JasonGross (approved):

.github/workflows/ci.yml
favonia: Good point.

favonia:
Let me merge this first, especially before #1106.

PR #1110, opened 2022-03-24
Concatenate errata-marking and PDF-making and remove DVI-making

favonia:
Previously, GitHub chose to check out the same commit hash in later jobs, breaking the assumption that the second job (PDF making) could pick up the commits pushed by the first job (errata marking). The second job thus failed because `errata.tex` does not appear to be clean. By concatenating two jobs into one, we no longer have to worry about carrying over the correct commit to make PDFs.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -115,7 +115,7 @@ log-check-for-warnings:
 	- ! grep -n "## Warning" hott-online.ilg /dev/null
 
 $(BOOKAUXFILES) : %.aux : %.tex
-	echo "WARNING: assuming $> is up-to-date"
+	echo "WARNING: assuming $@ is up-to-date"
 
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -115,7 +115,7 @@ log-check-for-warnings:
 	- ! grep -n "## Warning" hott-online.ilg /dev/null
 
 $(BOOKAUXFILES) : %.aux : %.tex
-	echo "WARNING: assuming $> is up-to-date"
+	echo "WARNING: assuming $@ is up-to-date"
```
favonia, line 118: @andrejbauer I believe this is a typo, right?

mikeshulman (approved):

Issue #1111, opened 2022-04-12
Code of conduct


andrejbauer:
Following the events of #1101, I am opening an issue to discuss the adoption of a code of conduct.

Please note that we already are bound by the [GitHub Community Forum Code of Conduct](https://docs.github.com/en/site-policy/github-terms/github-community-forum-code-of-conduct). The question is whether we should and want to have an additional Code, and if so, what? @atondwal made pertinent obsercations in [this comment](https://github.com/HoTT/book/pull/1101#issuecomment-1096447140).

I have very little exeperience with formulating Codes of conduct. As far as I am concerned the GitHub one that we already abide by is good enough, even though it may not be perfect in every respect.

Let me also say that I do not have the spare capacity to actively contrubite to formulation and adoption of the Code. If there are other book authors who are interested in heading the issue, I will be grateful to you.

EgbertRijke:
I think a code of conduct should be brief and focused on the book project. The book project is not a community project, but it does take input from the community. It should make people who want to give their input feel welcome, that their input will be valued, remind everyone on the importance of kindness and respect, it should state what can be expected of us book authors. It should also address specifically the topic of inclusive language, and state that we will use inclusive language including gender neutral pronouns.

The book is for many people their entrance to the homotopy type theory community and our field of research. So I think it is important for our community to have a welcoming statement, and that we commit to behaving accordingly.

In reaction to the discussion that followed upon the simple request to use inclusive language in the book, I also felt it is important to write a statement of inclusion for the `agda-unimath` project. This project is intended to be community driven, so that influences the statement, but anyway here it is:

https://github.com/UniMath/agda-unimath/blob/master/README.md#statement-of-inclusion

dlicata335:
I agree with Egbert.  The inclusivity statement I use on course syllabi is: 

> My goal is to welcome all students to computer science. I believe that everyone in the class is fully capable of mastering the material. Our classroom should be an inclusive space, where everyone is treated with respect, and ideas, questions, and misconceptions are discussed with respect. All students are encouraged to actively participate, regardless of gender, race, class, able-bodiedness, nationality, native language, sexual orientation, political ideology, or religious beliefs. All students are responsible for contributing to a welcoming environment in all course activities, such as labs, TA sessions, and online posts. If the course environment does not meet these standards, I would appreciate it if you would speak to me about it, so that I can do better in the future.

[...]

PR #1101, opened 2022-03-11
Use inclusive language in §1.11

ThePuzzlemaker:
Change "his or her" -> "their"

PR #1112, opened 2022-04-13
Code of conduct

EgbertRijke:
Following the suggestions in #1111 and #1101, I wrote an initial draft for a code of conduct. Any suggestions for improvements are much appreciated.


README.md
```diff
@@ -14,6 +14,14 @@ Compiled and printed versions of the book are available at the
 and nightly builds are available on the
 [github wiki](https://github.com/HoTT/book/wiki/Nightly-Builds).
 
+## Editing the book
+
+This book is not a community project, but we do welcome our readers to suggest minor improvements. The best way to propose an edit is to open a pull request with your suggested change.
```
favonia, line 19:
```suggestion
This book is not a community project, but we do welcome our readers to suggest improvements. The best way to propose an edit is to [open a pull request](https://github.com/HoTT/book/compare) with your suggested change. You can also [open an issue](https://github.com/HoTT/book/issues/new/choose) if you do not have a concrete proposal yet.
```

EgbertRijke:
Thanks for the edits Favonia! Is there a useful guideline for inclusive language that we could link to?

favonia:
> Is there a useful guideline for inclusive language that we could link to?

See my comment in #1111.

README.md
```diff
@@ -14,6 +14,14 @@ Compiled and printed versions of the book are available at the
...
+
+## Code of conduct
+
+For many, the HoTT book is their introduction to our subject and our diverse community, including people from any nationality, gender identity, sexual orientation, race, colour, ability, and background. In order to ensure for everyone a welcoming and inclusive environment in our discussions, we follow the guidelines of the [github code of conduct](https://docs.github.com/en/site-policy/github-terms/github-community-forum-code-of-conduct). You can expect from the authors and any participant that we are kind and respectful in discussions, that we use inclusive language, and that we do our best to understand each other's different perspectives. It might not always be that we accept the change in the way you proposed it, but we always value your input, regardless of your level of experience or status within the community.
```
atondwal: This is fairly large refactoring, but I think it's best to explicitly include what we care about from the github CoC (and perhaps cite it) rather than to invoke it by name. It's far too legalistic to simply #include, IMO. See, for example, how the [django coc](https://www.djangoproject.com/conduct/) evolved from the SpeakUp! coc.

FWIW I also think basing our CoC on one like Django's or Speakup's would be better as they are more focused on maintaining a healthy community, while GH's is mostly cover-your-ass. cf [this comment](https://github.com/HoTT/book/pull/1101#issuecomment-1096447140) for more on this.

favonia: @andrejbauer @atondwal @EgbertRijke I wonder if it makes more sense to have a different PR? I think we need a protocol for people to propose different CoCs, refine each of them separately, discuss the differences among them, and eventually pick one. Multiple PRs might be a way, and I wonder whether there's a better one. In any case, I am looking forward to the version proposed by @atondwal.

andrejbauer:
Would it be possible to get this PR resolved? I would be ok with accepting it now, and improving the Code of conduct later. Improving CoC until it becomes perfect is not going to lead to having one.

awodey:
Look on the conference website. We’ve already adopted the ASL CoC

Issue #1113, opened 2022-04-21
What qualifies as a "constructor" and the omitted $\equiv$ rules in appendix A.2


Ryunaq:
Disclaimer: I'm new to type theory.

I'm trying to have a good understanding of the formalism in appendix A.2. The book says the following

> Additionally, for all the type formers below, we assume rules stating that each constructor preserves definitional equality in each of its arguments; for instance, along with the Π-INTRO rule, (Π-INTRO-EQ rule here).

Given the example, I understand that there is a rule X-INTRO-EQ for every type former X. I would understand the same thing applies to make rules X-FORM-EQ. What I'm not sure is if the same applies for rules X-ELIM-EQ, since the eliminations rules do not seem to "construct" but rather eliminate. I've been told they do count, but I wanted to double check and maybe do a pull request specifying that this applies to exactly these 3 types of rules.

This arised when I tried to show that `f \equiv g` implies `f(a) \equiv g(a)` but I got stucked.

mikeshulman:
I think yes, this should apply to X-ELIM-EQ as well.  If so, I think you're right that the word "constructor" is confusing.   I think the intent was that we often view the syntax as inductively generated at the meta-level, and in this case each type- and term-forming rule (including both introductions and eliminations) is a meta-level "constructor" of this inductive type of syntax.  Would it be clearer to say something like "each type- and term-forming operator"?

Ryunaq:
I see, thanks for clearing the use of the word "constructor".  Yes, I think your proposed solution would work fine. Maybe just mention the full list for the Π-Types, to be on the safe side, I'll be making a PR.

cangiuli:
Yes, this applies also to elimination rules; there is a PI-ELIM-EQ.

FWIW I'd say that PI-ELIM "constructs" a term of type B[a/x], but I see how this phrasing is confusing. +1 to Mike's suggestion "each type- and term-forming operator".

PR #1114, opened 2022-04-21
Elaborate more on the omitted eq rules

Ryunaq:
Edited as mentioned in https://github.com/HoTT/book/issues/1113. I do fear that now lines 612 and 613 may confuse the meaning of "these local principles" in line 614 as to only refer to the dependent function type rules. I do not know how to express this better. Maybe those two lines should be removed after all. I leave this to your judgment. 
Thanks for maintaining this book :)

### DIFF
--- a/formal.tex
+++ b/formal.tex
@@ -599,8 +599,9 @@ assume that judgmental equality is an equivalence relation respected by typing.
 \end{mathparpagebreakable}
 %
 Finally, we assume that judgmental equality is a congruence respected by typing,
-i.e., that each constructor preserves judgmental equality in each of its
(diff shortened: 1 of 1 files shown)


cangiuli:
Do you think the passage is potentially confusing if lines 612-613 are deleted? Alternatively, you could edit line 614 to say something like "these local principles (at every type)".

Ryunaq:
I like the second alternative, I've comitted those changes too.

mikeshulman:
Looks good to me, thanks!  If no one has any further suggestions after a bit, we can merge it.

PR #1129, opened 2022-10-31
adjust references to origin of Martin-Löf identity types

ecavallo:
General identity types do not appear in [Martin-Lof-1972](https://archive-pml.github.io/martin-lof/pdfs/An-Intuitionistic-Theory-of-Types-1972.pdf) (though there is an equality type for natural numbers defined by recursion), but rather in [Martin-Lof-1973](https://archive-pml.github.io/martin-lof/pdfs/An-Intuitionistic-Theory-of-Types-Predicative-Part-1975.pdf).


awodey:
Thanks!

PR #1130, opened 2022-11-07
Inverse of a Dedekind real

valis:
Fixes the definition of an inverse of a Dedekind real


mikeshulman:
We discussed this on Zulip and it looks necessary and correct to me.  Any other opinions, e.g. @andrejbauer?

PR #1146, opened 2023-09-05
fix syntax of errata merge, close #1144

### LINKED ISSUE #1144: Errata PDF unreadable
The column spacing in the errata PDF (caused by an overly wide "Fixed in" entry) pushes explanations past the right edge of the page.

mikeshulman:


mikeshulman:
Hah, this is better, but apparently the name of that merge commit is very long, for some reason it starts with `public-release-` unlike any of the others.  I guess we fix that manually.

PR #1147, opened 2023-09-05
make commit name shorter

mikeshulman:


mikeshulman:
That's better.  Although it's still running a bit past the right-hand margin, maybe because the widths of the columns were set when the commit numbers were smaller?  Not sure if that's worth mucking with.

PR #1152, opened 2024-01-02
Fix Typos in categories.tex

niyoushanajmaei:


categories.tex
```diff
@@ -468,7 +468,7 @@ \section{Functors and transformations}
   \[ \xymatrix{
     G\circ (1_B \circ F) \ar@{=}[rr] \ar@{=}[dr] &&
     (G\circ 1_B)\circ F \ar@{=}[dl] \\
-    & G \circ F.}
+    & G \circ F}
```
mikeshulman, line 471: Wouldn't it be more consistent to keep the period here also, but change the preceding phrase to end with a colon?

PR #1154, opened 2024-02-24
Fix typos in the proof of Theorem 8.4.6

EricWay1024:
In the proof of Theorem 8.4.6,
- The order of function composition `g o f` as well as the truncated version should be reversed;
- The identity `g(t) = w'` should be `|g(t)|_0 = w'` so that both sides lie in `||W||_0`;
- `|fib_f (z_0)|_0` should be `||fib_f (z_0)||_0`.


mikeshulman:
Thank you!  Sorry for the delay in reviewing.

PR #1164, opened 2025-03-05
give equal credit to several originators of inductive schemas

mikeshulman:
Thierry Coquand suggested that this way of assigning credit for schemas for inductive types would be more fair.


awodey:
lgtm - thanks!

awodey:
BTW, Andrej just told me that the book has sold nearly 5000 copies with total revenue of about $20k!

awodey:
make that $100k!

PR #1171, opened 2025-05-06
Explicit counterexample where (x, u) = (x, v) but u ≠ v

jeanas:


mikeshulman:
This is a nice idea, but my inclination would be to delay the statement of the specific example until after univalence is formally introduced.  Either make it a "Remark 2.10.6" or an "Exercise 2.19", and refer forwards to that from Remark 2.7.1.  Anyone else ave an opinion?

awodey:
I agree - make it an exercise.

mikeshulman:
Thanks!

PR #1185, opened 2025-07-01
Fix typo: Use en dash for separating names: Escardó–Simpson instead of Escardó-Simpson 

Bolpat:


Bolpat:
Okay, I removed the errata commit.

Issue #1172, opened 2025-05-06
CI workflow broken


mikeshulman:
The [generate nightlies](https://github.com/HoTT/book/actions/runs/14865620471/job/41741570998) workflow is stopping with an error, but I'm having trouble figuring out what the error is.  Maybe I am blind but I don't see any errors other than undefined references that ought to be fixed by a second run.

mikeshulman:
Thanks for the suggestion.  I tried it, but it didn't help; the CI still errors.

enklht:
Thanks for giving it a try. It still errors, but according to the log, the failure now occurs in a different place: the main book compiles, and the error happens during the compilation of errata.tex.

mikeshulman:
Thanks again for this workaround.  I'm going to leave this issue open for now, though, in case anyone feels like diagnosing and fixing whatever the root cause may be.

PR #1187, opened 2025-07-21
Unvendor packages to fix CI

danielzgtg:
Something related to "fixcleverref" is incompatible with texlive-full=2024.20250309-1 on CI or Ubuntu 25.04. The CLI had "Latex failed to resolve 216 citation(s)". The end of the introduction had "??" instead of links, and the word "section" in unstyled text before headings.

Unvendoring all the (now-) unnecessarily vendored local packages fixes the problem. I enabled `-g` for `latexmk` to fix the .pdf not being rebuilt when I delete it.

# Before

?? and other breakage everywhere

# After

links work

Fixes: #1172


mikeshulman:
I don't remember why those packages were vendored in the first place.

danielzgtg:
a9eb3bcfd5e5828f824fa36db9d8e2a6b9550b06 added some because "The IAS machines and several other people seem to have old geometry packages installed."

Well, that was 12 years ago. The solutions are either to simply unvendor the packages whereby the breakages on CI and modern machines are fixed, or to have someone spend a lot of time finding the correct version of peer dependencies to also vendor.

mikeshulman:
Thanks for the sleuthing!

PR #1188, opened 2025-07-21
Unify em dashes to spaced version

danielzgtg:
Em dashes were annoyingly randomly spaced and unspaced. Edit: For mathematical writing, the spaced version was requested as clearer.


mikeshulman:
I know that most style guides recommend no spaces around em dashes, but I think there's an argument that in mathematical writing with inline math spaces are preferable, since an em dash adjacent to inline math without a space might be mistaken for part of the formula.  I think we might even have had an informal convention of using spaces around em dashes.  How many instances are there of em dashes without spaces?

danielzgtg:
> spaced em dashes in the rest of the book, so I'd be inclined to standardize on that.

I redid this to standardize on spaced em dashes as requested. The branch name is unchanged to avoid creating a new PR.

mikeshulman:
Thanks!

PR #1191, opened 2025-08-27
Remove big space

Bolpat:
I don’t know if this even works, but look for a bigger-than-usual space between “called the homotopy hypothesis/theorem,” and “because”.

**Edit:** There are many such cases. I don’t know if I have time to find and fix them all.


basics.tex
```diff
@@ -94,9 +94,9 @@ \chapter{Homotopy type theory}
 the fundamental $\infty$-groupoid construction is adjoint\index{adjoint!functor} to the
 geometric\index{geometric realization} realization of an $\infty$-groupoid as a space, and this
 adjunction preserves homotopy theory (this is called the \emph{homotopy
-  hypothesis/theorem},
-\index{hypothesis!homotopy}
-\index{homotopy!hypothesis}
+  hypothesis/theorem},%
```
mikeshulman, line 97:
```suggestion
  hypothesis/theorem},
```

